Issue #1, opened 2015-09-18
Asynchronous client callbacks may not be executed sequentially


kuujo:
The current implementation provides sequential consistency for asynchronous clients by sequencing commands as they're applied to the Raft state machine. If a client submits a command through two different paths, client sequence numbers instruct the internal state machine on how to resequence writes such that if a client submits multiple concurrent writes, they'll be applied in the order in which they were sent. 

However, this algorithm does not currently preserve sequential consistency in responses. That is, if a client submits multiple commands concurrently, even if the commands are applied to server state machines in the correct order, the client's `CompletableFuture`s may still be completed out of order. This can happen in the event that a response is lost on the way to the client. In that case, the client will resubmit the command, but any responses received thereafter will be applied upon receipt and may therefore be completed out of order.

In order to resolve this issue, clients need to sequence command responses. This issue also applies to queries. From a client's perspective, it can see reads go back in time when multiple reads are executed concurrently. Clients simply have to queue responses received out of order. We may want to provide options for enabling/disabling client linearizability. Sequentail consistency algorithms apply only to asynchronous usage of the client since synchronous usage enforces sequential consistency by its nature.

PR #9, opened 2015-09-21
Handle sequential execution of asynchronous callbacks and improve client consistency models

kuujo:
This PR resolves issue #1 and extends the concept of consistency levels in Catalog. With this PR, Catalog supports two consistency levels for commands and four for queries. Following are the proposed consistency levels. Note that while consistency levels allow the client to specify order of execution for commands and queries, all concurrent asynchronous callbacks are executed sequentially in all cases.

**Command consistency levels**
* `CAUSAL` - The causal consistency level guarantees consistency between non-overlapping commands from a single client. Essentially, what this means is that the internal state machine does not attempt to reorder causally consistent commands. Because the nature of Raft is such that once a response is received, the command is guaranteed to have been committed, simply relaxing the order requirement for concurrent requests from a single client gets us causal consistency from the perspective of that client. Note that this only applies to asynchronous usage.
* `LINEARIZABLE` - The linearizable consistency level guarantees sequential consistency for all writes from a client. That is, if a client writes `a`, `b`, and `c`, in that order, those commands will be applied to the state machine in that order and responses will be received in that order. Because writes are linearizable, this sequential consistency makes correct linearizability possible.

**Query consistency levels**
* `CAUSAL` - Causal consistency requires that non-overlapping reads be sequential, but concurrent reads can be reordered. How this is implemented is by sending the client's last seen index and the last successful write from the client with each request. Server state machines will ensure state progresses at least up to the last write or the last index seen by the client, whichever is greater.
* `SEQUENTIAL` - Sequential consistency requires that all reads - including concurrent ones - see state progress monotonically. In some cases, queries can be reordered if the client switches servers during concurrent executions, so clients must ensure order is preserved in responses. This works similarly to causal consistency - by sending the client's last seen index - with the additional constraint that if a client receives an ordered response for a version prior to the last seen version, it will resubmit the request. In practice, except in failure cases, clients should usually see responses in sequential order because of the nature of state machines.
* `BOUNDED_LINEARIZABLE` - This is the same as the `LINEARIZABLE_LEASE` consistency level which existed previously. Queries will be forwarded to the leader, and the leader will apply the query to its own state machine, assuming that it must have contacted a majority of the cluster within the last heartbeat timeout (leaders will step down if they fail to contact a majority of the cluster).
* `LINEARIZABLE` - This is the same `LINEARIZABLE` consistency level as previously existed. Linearizable queries will go to the leader, and the leader will synchronously verify its leadership with a majority of the cluster before responding.


kuujo:
This actually still needs some work. In order to maintain causal consistency for writes, we still need to ensure they can only be applied once to the state machine. However, the allowance of out-of-order commands inherent to causal consistency makes tracking state machine output difficult. This implementation does not include sequence numbers for `CAUSAL` commands.

However, even though causal commands may be logged and applied to the state machine out-of-order, clients still ensure that their responses Re handled in the order in which they were sent, and that means clients can still reliably ack responses in sequence order. Clients should send sequence numbers with all commands and ack commands in normal order *after* responses are resequenced on the client.

kuujo:
Moving this to #11 to combine the two issues due to the previously mentioned remaining bug.

PR #11, opened 2015-09-21
Sequence commands prior to logging

kuujo:
This PR extends #9 to move sequencing of `LINEARIZABLE` commands to the Raft leader. Commands must be sequenced prior to writing to the log, otherwise commands that are cleaned from the log will represent breaks in the sequencing of a set of commands for a specific session, and the state machine will not be able to advance. With the leader sequencing commands to the Raft log, state machines no longer have to assume commands may have been written out of order.


kuujo:
I've merged #9 with this PR. Here's the outline of the earlier changes from that PR:

With this PR, Catalog supports two consistency levels for commands and four for queries. Following are the proposed consistency levels. Note that while consistency levels allow the client to specify order of execution for commands and queries, all concurrent asynchronous callbacks are executed sequentially in all cases.

**Command consistency levels**
* `CAUSAL` - The causal consistency level guarantees consistency between non-overlapping commands from a single client. Essentially, what this means is that the internal state machine does not attempt to reorder causally consistent commands. Because the nature of Raft is such that once a response is received, the command is guaranteed to have been committed, simply relaxing the order requirement for concurrent requests from a single client gets us causal consistency from the perspective of that client. Note that this only applies to asynchronous usage.
* `LINEARIZABLE` - The linearizable consistency level guarantees sequential consistency for all writes from a client. That is, if a client writes `a`, `b`, and `c`, in that order, those commands will be applied to the state machine in that order and responses will be received in that order. Because writes are linearizable, this sequential consistency makes correct linearizability possible.

**Query consistency levels**
* `CAUSAL` - Causal consistency requires that non-overlapping reads be sequential, but concurrent reads can be reordered. How this is implemented is by sending the client's last seen index and the last successful write from the client with each request. Server state machines will ensure state progresses at least up to the last write or the last index seen by the client, whichever is greater.
* `SEQUENTIAL` - Sequential consistency requires that all reads - including concurrent ones - see state progress monotonically. In some cases, queries can be reordered if the client switches servers during concurrent executions, so clients must ensure order is preserved in responses. This works similarly to causal consistency - by sending the client's last seen index - with the additional constraint that if a client receives an ordered response for a version prior to the last seen version, it will resubmit the request. In practice, except in failure cases, clients should usually see responses in sequential order because of the nature of state machines.
* `BOUNDED_LINEARIZABLE` - This is the same as the `LINEARIZABLE_LEASE` consistency level which existed previously. Queries will be forwarded to the leader, and the leader will apply the query to its own state machine, assuming that it must have contacted a majority of the cluster within the last heartbeat timeout (leaders will step down if they fail to contact a majority of the cluster).
* `LINEARIZABLE` - This is the same `LINEARIZABLE` consistency level as previously existed. Linearizable queries will go to the leader, and the leader will synchronously verify its leadership with a majority of the cluster before responding.

PR #18, opened 2015-09-28
Fix how batch of entries to replicate to follower is computed

madjam:
When the leader attempts to replicate a batch of entries to a follower, it currently attempts to populate the batch until one of two conditions are met
1. batch size is greater than MAX_BATCH_SIZE
2. we are at the end of the log before we breached the batch size limit

Existing logic for populating the batch can lead to some entries being skipped over.
Say for example batch size limit is 100 and we are currently at 80. If the size of next 2 entries in the log is 30 and 20 bytes respectively, we will end up skipping the 30 byte entry but include the subsequent 20 byte entry.

This patch attempts to fix it.

An open question is: can the entry is the log be null? The fix assumes that entries cannot be null. May be that isn't true.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -21,6 +21,7 @@ import io.atomix.copycat.client.error.RaftError;
 import io.atomix.copycat.client.error.RaftException;
 import io.atomix.copycat.server.CopycatServer;
 import io.atomix.catalyst.transport.Address;
+import io.atomix.catalyst.util.Assert;
 import io.atomix.catalyst.util.concurrent.ComposableFuture;
 import io.atomix.catalyst.util.concurrent.Scheduled;
 import io.atomix.copycat.client.request.*;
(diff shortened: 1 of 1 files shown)


kuujo:
@madjam I made some small modifications to your PR, but I wanted to make sure it got in. This actually helped me discover a remaining bug, so thanks. The modification I made is this: Copycat's implementation of the Raft protocol allows `null` entries to be replicated. Entries within the bounds of the log can be `null` if they've been compacted from the log. See the [log compaction API documentation](http://atomix.io/copycat/api/latest/io/atomix/copycat/server/storage/compaction/package-frame.html) and the classes therein for extensive documentation on how/why this is done.

`null` entries are effectively skipped by simply not adding them to an `AppendRequest`. This is safe and does not conflict with the safety properties of the Raft consensus algorithm since only committed entries are compacted from the log. Servers are always required to maintain the last committed entry in their log for consistency checks.

So, the modification I made was to remove the assertion and allow `null` entries by effectively skipping them. I had taken a stab at fixing this issue previously, but your PR actually made me realize it still wasn't handling `null` entries properly since my code was `break`ing on the first `null`.

PR #19, opened 2015-10-19
Cleanup commitFutures that have failed.

madjam:
Scenario: Start a 3 node cluster c1, c2 and c3. Let's assume c1 becomes LEADER. Shut down c2 and c3.
Expected: c1 should eventually detect that it is partitioned away and should step down and become a FOLLOWER

What we see is that c1 remains in leader state. Stops heart beating because commitFuture and nextCommitFuture are both non-null and incomplete. We never fail commitFuture when we detect the communication with follower(s) failed.

Later if we bring back c2 and c3 up, They will start their own polls since they can no longer receive heartbeats from c1. c1 continues to remain in LEADER state and can never rejoin others.

Solution: Exceptionally complete commitFuture if a quorum of successful commits cannot be guaranteed.
This change also disables exponential backoff when interacting with down followers as it can interfere with the ability to detect and fail commitFutures that have no chance of success.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -794,6 +794,7 @@ final class LeaderState extends ActiveState {
   private class Replicator {
     private final Set<MemberState> committing = new HashSet<>();
     private long commitTime;
+    private int commitFailures;
     private CompletableFuture<Long> commitFuture;
     private CompletableFuture<Long> nextCommitFuture;
     private final TreeMap<Long, CompletableFuture<Long>> commitFutures = new TreeMap<>();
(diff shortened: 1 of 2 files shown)


kuujo:
That test failure is unrelated.

kuujo:
This is definitely an issue. There is a bit of work to be done on the leader's logic with all the concurrency. I think this is probably an elegant enough solution. But I want to see how it does in Jepsen before it can be merged.

kuujo:
My initial tests show that this does indeed fix the leader partition detection. I must have broken it when I added exponential backoff. I think this will do fine in Jepsen, but will still verify that's the case anyways.

While this issue fixes the partition detection/step down, I still rather dislike that without exponential backoff it unnecessarily frequently connects/fails to a down follower. A single downed node is more common than a partition wherein the leader is on the minority side, and in that case the leader will remain the leader and end up doing a lot of unnecessary work attempting to repeatedly connect to the down follower.

Ideally, I think if the leader fails to contact a *majority* of the cluster perhaps it should forgo exponential backoff and figure out if it should step down, but in cases where the leader is able to maintain contact with a majority of the cluster, exponential backoff remains a useful tool as long as the maximum retry interval is not too great to find a follower once it restarted.

kuujo:
Alternatively, I suppose attempting to connect to a down follower at the rate of normal heartbeats is fine, but in that case we should probably include logic for minimizing the logging of failed connections to dead followers. I think that would make this PR acceptable.

Can we do something like log the first and `n%` of connection failures?

kuujo:
I verified that the tests pass aside from the unrelated test hanging above.

kuujo:
@madjam I decided to go ahead and implement limited logging for this PR. The logging logic does something like:
```java
if (failureCount <= 3 || failureCount % 100 == 0) {
  LOGGER.warn(...);
}
```

madjam:
@kuujo The logging change looks good to me. I too found the excessive logging a little annoying.

madjam:
I will submit another PR for a related fix that will ensure we only send non-empty append requests to followers that are up. This should fix a general slowdown I noticed when a down follower's log is way behind the leader.

PR #21, opened 2015-10-19
Fix: Cluster configuration local to ServerState can get out of sync with actual cluster configuration

madjam:
When cluster configuration changes (i.e. nodes join / leave) ServerState can get out of sync. For example: A node can reject polls from candidates that were not part of the original bootstrap configuration indicating that the node is unknown.

Fix: All cluster membership queries get posed against the the current cluster configuration. The lookup as implemented in this fix is O(n). But I felt that is reasonable given the expected cluster sizes. Alternatively a more efficient implementation is to use a listener for cluster changes.


kuujo:
Ahh makes sense. Good catch!

Need tests for this, but I can't judge because we need tests in general for configuration changes right now. TODO

PR #22, opened 2015-10-20
Only send emptyCommits to a down follower

madjam:
When a down follower's log is significantly behind the leader's log, each AE request from leader to this follower will be a full batch of log entries. This results in a lot of wasted work. With this change we only send non-empty AE request if the follower has successfully responded to recent communication.


madjam:
@kuujo: Any thoughts on this one?

kuujo:
Yeah I was just looking at it. Great optimization. I think I was too tired to realize what was going on when I looked at it at 2am the other night :-P Will merge momentito.

PR #28, opened 2015-10-22
Reject concurrent configuration changes

### LINKED ISSUE #26: Reject configuration changes if a configuration change is ongoing
The current implementation of configuration changes does not reject configuration requests if an existing reconfiguration is already taking place. This is a safety violation since multiple concurrent configuration changes can result in non-overlapping majorities. Leaders should ensure that configuration changes are committed and therefore applied on all servers before accepting new configuration changes.

kuujo:
This PR fixes #26. Both `JoinRequest`s and `LeaveRequest`s received by the leader will be rejected if the leader is already in the process of committing another configuration. The `ServerState`s `join()` and `leave()` methods will retry configuration changes indefinitely if a leader responds with a `null` error response (not very elegant but works until better server-side exceptions are added).

§[4.1](https://ramcloud.stanford.edu/~ongaro/thesis.pdf)

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -45,7 +45,7 @@ final class LeaderState extends ActiveState {
   private Scheduled currentTimer;
   private final Replicator replicator = new Replicator();
   private long leaderIndex;
-  private boolean configuring;
+  private long configuring;
 
   public LeaderState(ServerState context) {
     super(context);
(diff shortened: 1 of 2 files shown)


kuujo:
Actually, this safety constraint is most essential when the quorum of voting members changes, which is when a member is promoted from `PASSIVE` to `ACTIVE` or removed from the cluster. I'll opt to keep the `configuring` flag in all cases of configuration changes just to be safe, but I don't think it's actually necessary when adding a server to the `PASSIVE` members list since the quorum does not change.

kuujo:
Okay, this PR was updated to include checks to ensure promoting members from `PASSIVE` to `ACTIVE` is done safely as well. Essentially, when a configuration change is started, the index of the configuration entry is stored in the `LeadsrState`. If another configuration change is requested while `configuring` is non-zero, it will be rejected. Storing the configuration entry index allows that index's `CompletaeFuture` to be referenced to wait for a configuration change to be completed for promoting members.

PR #31, opened 2015-10-23
Allow servers to compact their logs once all events have been acked

### LINKED ISSUE #25: Allow log compaction for all committed entries
Currently, logs are not allowed to be compacted until entries have been stored on active all servers in the cluster. This ensures that all state machines see every command and thus all state machines store related events in memory until they're acked by client sessions. But this is obviously not ideal because, essentially, this means logs cannot be compacted while any active server is down. Ideally, servers should be allowed to compact entries that have been committed.

The issue with compacting entries that haven't been stored on all servers is that if a server compacts a command that triggers an event before that command is received by some follower, the state of events on all servers will become inconsistent. Sessions ensure they receive events in sequential order by checking each event message against the previous received event message for consistency. Consider the following scenario:
* In a three node cluster with servers `A`, `B`, and `C`, server `A` is the leader
* A client connects to server `C` and registers a session
* Server `C` is far behind the leader, but because session operations are proxied to the leader, the client cannot detect that server `C` is slow
* Another client submits a command that triggers a sequential event to be sent to the first session and the command is committed through servers `A` and `B`
* Server `A` compacts the command out of its log before server `C` catches up

In this scenario, server `C` will never receive the command and thus will never publish the event. That results in an inconsistent event history between server `C` and servers `A` and `B`.

An inconsistent history of events between servers is a safety violation. Clients should always see every event sent to their session, but if a client were connected to server `C` it would miss the event.

But in addition to safety, liveness also becomes an issue with inconsistencies between servers. Servers and clients coordinate sequencing of events in a manner similar to Raft's AppendEntries RPCs wherein the server sends the previous event index for comparison. However, previous event indexes for servers `A` and `B` will differ from that of `C`. A client may ack an event index, thus clearing events from server memory, and then switch to a server that expects the client to have seen events that it never saw. Essentially, a client may ack events that it doesn't know about.

Fortunately, Copycat already does keep track of which clients are connected to which servers through connect requests/responses. My proposed fix to this is to extend that mechanism to allow servers to reject connections from clients if conditions indicate that sequential events sent by that server may not be complete. Essentially, what we want is to ensure that any server to which a client is connected has the complete history of events from the client's last received event to the last committed index in the log. In order to do this, servers simply need to track the *lowest* index for which they've received sequential commits. When a client connects to a server, it sends the last received event index for its session. If the server's lowest sequential commit index is less than the client's highest event index, it rejects the connection and forces the client to find another server.

Because of the nature of consensus, clients should always be able to find a healthy server to which to connect and from which to receive events. That is, there will always be at least one server that participated in the commitment of each command, and so there will always be some server that received each command. Clients simply need to find the server that has a complete event history starting from their last received event.

In order to ensure entries are not compacted *after* a connection between a client and server is made, servers will have to use the lowest index of all connected clients in the same way that `globalIndex` is currently used. That is, servers will prevent entries from being compacted so long as some client connected to the server is awaiting events.

That last point is actually similar to an alternative solution. Since client keep alive requests are logged and replicated, it could be possible to use that metadata in place of `globalIndex` in general to ensure logs are never compacted until al open sessions have received events. Essentially, the *lowest* event index acked by any client becomes the `globalIndex`. This is certainly a much simpler solution, but one of the benefits to forcing clients to switch servers is that it forces clients to switch to more up-to-date servers.

kuujo:
This PR modifies the way logs are committed/compacted to fix #25.

Currently, Copycat does not allow servers to compact any segment that contains entries that haven't been stored on `100%` of the servers in the cluster. The reasoning behind this decision was because session events are dependent on commands being applied to the state machine. When events are published to a session, they're held in memory on each server until received by the client. Events are sequenced in a manner similar to how the leader sequences entries to followers. When an event batch is sent to a client, the server that sends the batch with the index of the previous event. The client verifies that it received the previous event and acks the new event in the response. While it's normally safe for a server to take a snapshot before other servers have applied a command, the dependency of events on commands means if one server doesn't apply *all* commands for each session, servers can end up with inconsistent event histories. Thus, in order to avoid this case, Copycat currently requires that a command be persisted on all servers before it can be removed from the log.

But this approach is not ideal for obvious reasons. As long as at least one server in a configuration is down, no server can compact its logs. This PR proposes a fix to allow servers to compact their logs once all events have been received by their respective clients.

Servers already track which events have been received by which clients through keep-alive requests. When a client sends a `KeepAliveRequest` for an open session, it sends with it the index of the highest event for which it has received a batch. Event indexes are based on the commands by which events were triggered. For instance, if the application of a command at index `100` triggers the publishing of an event to 5 sessions, the `index` of that event will be `100`. Once each session has received event `100` and sends a `KeepAliveRequest`, a `KeepAliveEntry` with the highest index received by the client is logged, replicated, committed, and applied to the internal state machine.

So, since the system already logs and replicates information about the highest event index received for each session, and because event indexes are based on log indexes, we can use the *lowest* acked event index for all sessions as the basis for safe log compaction. When a client submits a `KeepAliveRequest` and an associated `KeepAliveEntry` is committed and applied, each server recalculates the lowest index received by any session. Session IDs are initialized to the index of the session's `RegisterEntry`, and so each session's event index is also initialized to that index as well (by definition, sessions can't receive events prior to their registration). This ensures that the lowest received index is not reset to `0` when a new session is registered. When the lowest index received by all sessions is recalculated, the log is updated to make segments containing earlier entries available for compaction.

### DIFF
--- a/client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
+++ b/client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
@@ -695,19 +695,32 @@ public class ClientSession implements Session, Managed<Session> {
    */
   @SuppressWarnings("unchecked")
   private CompletableFuture<PublishResponse> handlePublish(PublishRequest request) {
-    if (request.session() != id)
-      return Futures.exceptionalFuture(new UnknownSessionException("incorrect session ID"));
+    LOGGER.debug("{} - Received {}", id, request);
 
+    // If the request is for another session ID, this may be a session that was previously opened
+    // for this client.
+    if (request.session() != id) {
+      LOGGER.debug("{} - Inconsistent session ID: {}", id, request.session());
+      return Futures.exceptionalFuture(new UnknownSessionException("incorrect session ID"));
+    }
+
+    // If the request's previous event version doesn't equal the previous received event version,
+    // respond with an undefined error and the last version received. This will cause the cluster
+    // to resend events starting at eventVersion + 1.
     if (request.previousVersion() != eventVersion) {
+      LOGGER.debug("{} - Inconsistent event version: {}", id, request.previousVersion());
       return CompletableFuture.completedFuture(PublishResponse.builder()
         .withStatus(Response.Status.ERROR)
-        .withError(RaftError.Type.INTERNAL_ERROR)
         .withVersion(eventVersion)
         .build());
     }
 
+    // Store the event version. This will be used to verify that events are received in sequential order.
     eventVersion = request.eventVersion();
 
+    // For each event in the events batch, call the appropriate event listener and create a CompletableFuture
+    // to be called once the event callback is complete. Futures will ensure that an event is not acknowledged
+    // until all event callbacks have completed.
     List<CompletableFuture<Void>> futures = new ArrayList<>(request.events().size());
     for (Event<?> event : request.events()) {
       Listeners<Object> listeners = eventListeners.get(event.name());
@@ -716,9 +729,15 @@ public class ClientSession implements Session, Managed<Session> {
       }
     }
 
+    // Wait for all event listeners to complete and then respond to the event message. This ensures that
+    // linearizable events are completed between their invocation and response. If the async queue is backed
+    // up and we don't wait for callbacks to complete, the cluster will believe an event to have been received
+    // and handled before it has indeed been received and handled.
     return CompletableFuture.allOf(futures.toArray(new CompletableFuture<?>[futures.size()]))
       .handleAsync((result, error) -> {
+        // Store the highest version for which event callbacks have completed.
         completeVersion = Math.max(completeVersion, request.eventVersion());
+
         return PublishResponse.builder()
           .withStatus(Response.Status.OK)
           .withVersion(eventVersion)
--- a/client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
+++ b/client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
@@ -639,6 +639,7 @@ public class ClientSession implements Session, Managed<Session> {
   private void onOpen(long sessionId) {
     LOGGER.debug("Registered session: {}", sessionId);
     this.id = sessionId;
+    this.eventVersion = id;
     this.state = State.OPEN;
     for (Consumer<Session> listener : openListeners) {
       listener.accept(this);
--- a/protocol/src/main/java/io/atomix/copycat/client/response/PublishResponse.java
+++ b/protocol/src/main/java/io/atomix/copycat/client/response/PublishResponse.java
@@ -68,20 +68,23 @@ public class PublishResponse extends SessionResponse<PublishResponse> {
     status = Status.forId(buffer.readByte());
     if (status == Status.OK) {
       error = null;
-      version = buffer.readLong();
-    } else {
+    } else if (buffer.readBoolean()) {
       error = RaftError.forId(buffer.readByte());
     }
+    version = buffer.readLong();
   }
 
   @Override
   public void writeObject(BufferOutput<?> buffer, Serializer serializer) {
     buffer.writeByte(status.id());
-    if (status == Status.OK) {
-      buffer.writeLong(version);
-    } else {
-      buffer.writeByte(error.id());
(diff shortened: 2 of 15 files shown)


kuujo:
So, there's actually still a problem with this algorithm. Consider the following scenario:
* A client registers a session at index `50` in the log
* The client's session is initialized with session ID `50` and event index `50`
* A server applies a command at index `100` which publishes event `100` to the client
* The client acks event index `100`
* The server then compacts index `100` from its log and crashes before applying any more commands or publishing any more events
* The server recovers and replays its log up to index `100`
* The client reconnects to the server
* The server then applies a command at index `101` which publishes another event to the client's session
* That event will be published with index `101` and previous index `50` (the session's initial event index), while the client's previous index will still be `100`

There are a couple ways this could be handled. First, servers could keep only compact logs up to the *previous* command for which an event was published to the client. However, the downside to this is that a session that receives events infrequently could keep the log from being compacted. If a session receives one event over its entire lifetime, logs can't be compacted until the session is unregistered. In that case, the `globalIndex` could be reintroduced to compact commands that are stored on all servers.

But a better way to do this might be to use the client's `KeepAliveEntry` event index as the previous index if the event index is greater than the previous event index for the server's session state. In other words, in the scenario above, since the client acked event `100` in a keep-alive request, that index can be used as the previous index for the event. Safety: because commands will never be compacted until a `KeepAliveEntry` with a given index has been logged for *all* sessions, commands between any client's last `KeepAliveRequest` and the end of the log should *never* be missing.

There's also a liveness issue here with respect to log compaction. If log compaction is dependent on clients acking events received via their session, that implies that a client must *have* events to ack. However, this may frequently not be the case. Some mechanism is necessary to ensure that log compaction advances for sessions that don't receive event messages or infrequently receive events. If a client never receives published events, how can it acknowledge that it knows about an event index?

To resolve this final issue, I propose using information about the event history *following* the last event received by a client. Once a client acks an event at index `i`, it essentially acknowledges receipt of all events up to the next published event. So, in the scenario above, once the client acks event `101`, if the next message published is at index `201`, then the client effectively acknowledged receipt of events `102-200` as well. If no events have been published to a session since the last acked event index, then the event index for that session is equivalent to `lastApplied`.

kuujo:
So, there's actually still a problem with this algorithm. Consider the following scenario:
* A client registers a session at index `50` in the log
* The client's session is initialized with session ID `50` and event index `50`
* A server applies a command at index `100` which publishes event `100` to the client
* The client acks event index `100`
* The server then compacts index `100` from its log and crashes before applying any more commands or publishing any more events
* The server recovers and replays its log up to index `100`
* The client reconnects to the server
* The server then applies a command at index `101` which publishes another event to the client's session
* That event will be published with index `101` and previous index `50` (the session's initial event index), while the client's previous index will still be `100`

There are a couple ways this could be handled. First, servers could keep only compact logs up to the *previous* command for which an event was published to the client. However, the downside to this is that a session that receives events infrequently could keep the log from being compacted. If a session receives one event over its entire lifetime, logs can't be compacted until the session is unregistered. In that case, the `globalIndex` could be reintroduced to compact commands that are stored on all servers.

But a better way to do this might be to use the client's `KeepAliveEntry` event index as the previous index if the event index is greater than the previous event index for the server's session state. In other words, in the scenario above, since the client acked event `100` in a keep-alive request, that index can be used as the previous index for the event. Safety: because commands will never be compacted until a `KeepAliveEntry` with a given index has been logged for *all* sessions, commands between any client's last `KeepAliveRequest` and the end of the log should *never* be missing.

There's also a liveness issue here with respect to log compaction. If log compaction is dependent on clients acking events received via their session, that implies that a client must *have* events to ack. However, this may frequently not be the case. Some mechanism is necessary to ensure that log compaction advances for sessions that don't receive event messages or infrequently receive events. If a client never receives published events, how can it acknowledge that it knows about an event index?

To resolve this final issue, I propose using information about the event history *following* the last event received by a client. Once a client acks an event at index `i`, it essentially acknowledges receipt of all events up to the next published event. So, in the scenario above, once the client acks event `101`, if the next message published is at index `201`, then the client effectively acknowledged receipt of events `102-200` as well. If no events have been published to a session since the last acked event index, then the event index for that session is equivalent to `lastApplied`.

Issue #29, opened 2015-10-22
ISE: inconsistent index


madjam:
I haven't done much debugging on this yet. But thought I will log this first.

Here's what I was doing:
- 3 node cluster
- A separate client is submitting commands to the cluster at the rate of few hundred per second
- Randomly kill a node bring it back up and verify everything is still running ok.

Here's what I did to see this error:
- Kill the leader
- Noticed another node was elected as new leader
- This expection is logged on the new leader

`12:21:00.424 [copycat-server-Address[localhost/127.0.0.1:5003]] ERROR i.a.c.u.c.SingleThreadContext - An uncaught exception occurred
java.lang.IllegalStateException: inconsistent index: 1605633
	at io.atomix.catalyst.util.Assert.state(Assert.java:69) ~[classes/:na]
	at io.atomix.copycat.server.storage.Segment.get(Segment.java:319) ~[classes/:na]
	at io.atomix.copycat.server.storage.Log.get(Log.java:319) ~[classes/:na]
	at io.atomix.copycat.server.state.LeaderState$Replicator.entriesCommit(LeaderState.java:1040) ~[classes/:na]
	at io.atomix.copycat.server.state.LeaderState$Replicator.commit(LeaderState.java:980) ~[classes/:na]
	at io.atomix.copycat.server.state.LeaderState$Replicator.lambda$commit$126(LeaderState.java:876) ~[classes/:na]
	at io.atomix.copycat.server.state.LeaderState$Replicator$$Lambda$111/982607974.apply(Unknown Source) ~[na:na]
	at java.util.Map.computeIfAbsent(Map.java:957) ~[na:1.8.0_25]
	at io.atomix.copycat.server.state.LeaderState$Replicator.commit(LeaderState.java:874) ~[classes/:na]
	at io.atomix.copycat.server.state.LeaderState$Replicator.access$100(LeaderState.java:818) ~[classes/:na]
	at io.atomix.copycat.server.state.LeaderState.accept(LeaderState.java:652) ~[classes/:na]
	at io.atomix.copycat.server.state.ServerState.lambda$registerHandlers$20(ServerState.java:421) ~[classes/:na]
	at io.atomix.copycat.server.state.ServerState$$Lambda$34/1604755402.handle(Unknown Source) ~[na:na]
	at io.atomix.catalyst.transport.NettyConnection.handleRequest(NettyConnection.java:102) ~[classes/:na]
[...]

PR #33, opened 2015-10-24
Reset all member state after transitioning to leader

kuujo:
This PR ensures that *all* member state is reset (aside from the `index`) when a server transitions to `LEADER`. See the discussion in #29 

@madjam can you check this out?


madjam:
This looks good to me.

PR #43, opened 2015-10-28
Ensure events published during session registration are associated with register entries

### LINKED ISSUE #41: State machines should be able to publish events during session registration
In order to ensure consistency, the internal `ServerStateMachine` only allows events to be published during certain operations. Effectively, for fault tolerance session events may only be published in response to operations that are replicated through the Raft log and are therefore applied on a majority of servers. But the current implementation is missing support for publishing session events during session registration operations. Session events published during session registration should be linearizable by default.

kuujo:
This PR fixes #41 by simply setting the context prior to calling the state machine's `register` method. This will ensure that events published in the `register` method are properly associated with the index of the `RegisterEntry` and that events will be received by the client before the session registration is complete.

One additional safety measure added in this PR is to ensure that state machines cannot publish events to the session being registered. If events are published to the session being registered, events would never be acknowledged by the client since it doesn't yet know about its session. So, we throw `IllegalStateException` on attempts to publish events to the session being registered, but publishing to other sessions should be allowed and events will be linearizable.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -603,7 +603,9 @@ final class LeaderState extends ActiveState {
       if (isOpen()) {
         if (commitError == null) {
           RegisterEntry entry = context.getLog().get(index);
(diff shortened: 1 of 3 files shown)


kuujo:
#48 fixed the tests

PR #44, opened 2015-10-28
Ensure timestamps in the internal state machine progress monotonically

### LINKED ISSUE #40: The view of time in the internal state machine may not be consistent
The internal `ServerStateMachine` tracks time by updating a timestamp with `max(entry.getTimestamp(), currentTimestamp)`. But in some cases, the entry's timestamp is used directly to evaluate things like session expiration:
https://github.com/atomix/copycat/blob/master/server/src/main/java/io/atomix/copycat/server/state/ServerStateMachine.java#L200

This can result in unexpected behavior in the state machine particularly after leadership changes. All state machine methods should store and use the same timestamp for time-dependent operations in order to ensure safety and consistency.

kuujo:
This PR fixes #40 to ensure time in state machines is managed consistently. Essentially, the management of time is centralized in `ServerStateMachineExecutor`. Each internal state machine method that uses time gets the time by calling `tick` on the state machine executor. `tick` does `max(timestamp, currentTimestamp)` to ensure that time progresses monotonically.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/ServerStateMachine.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/ServerStateMachine.java
@@ -80,14 +80,20 @@ class ServerStateMachine implements AutoCloseable {
 
   /**
    * Sets the last applied index.
+   * <p>
+   * The last applied index is updated *after* each time a non-query entry is applied to the state machine.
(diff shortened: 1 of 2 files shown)


kuujo:
#48 fixed the tests

PR #45, opened 2015-10-28
Prevent race conditions when expiring sessions

### LINKED ISSUE #42: Expired sessions may not be expired in the internal state machine in certain cases
There's a race condition in the way sessions are currently expired. When register, keep alive, and unregister entries are applied to the state machine on each server, timestamps for all sessions are checked and sessions are marked as suspicious if their expiration time has passed. But the leader is responsible for expiring sessions that are marked suspicious. This ensures that sessions cannot be expired during leadership changes or other events that prevent a client from keeping its session alive.

But there is a case where a client may submit a keep alive request - thus renewing its session timeout and marking its session as no longer suspicious - and a leader may log an `UnregisterEntry` while the client's keep-alive is in the process of being committed. In that case, the keep-alive will be committed and applied before the leader's `UnregisterEntry`, and the following logic will result in `close(Session)` but not `expire(Session)` being called on the state machine:
https://github.com/atomix/copycat/blob/master/server/src/main/java/io/atomix/copycat/server/state/ServerStateMachine.java#L302

The `UnregisterEntry` should be enriched with an `expired` flag to indicate whether the leader expired the session or the client explicitly closed its session. This will ensure that the proper state machine methods are called upon session expiration.

kuujo:
This PR fixes #42 by adding an `expired` field to `UnregisterEntry`. When a leader detects that a session has expired, it logs the `UnregisterEntry` with `expired=true`. This allows the internal state machine on each server to determine that the leader expired the session (as opposed to the client unregistering it).

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -171,6 +171,7 @@ final class LeaderState extends ActiveState {
         try (UnregisterEntry entry = context.getLog().create(UnregisterEntry.class)) {
           entry.setTerm(term)
             .setSession(session.id())
+            .setExpired(true)
             .setTimestamp(System.currentTimeMillis());
           index = context.getLog().append(entry);
           LOGGER.debug("{} - Appended {} to log at index {}", context.getAddress(), entry, index);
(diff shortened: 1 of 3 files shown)


kuujo:
#48 fixed the tests

PR #46, opened 2015-10-28
Ensure tombstones are retained in the log until applied on all servers

### LINKED ISSUE #38: Ensure tombstones are applied on all servers before being removed from the log
The changes in #31 removed the use of `globalIndex` to indicate when it was safe to compact server logs. Instead, they were replaced with the highest index for which events have been received by all clients. But it turns out that while this did correctly address the issue of log compaction while servers were down, it introduced a bug into compaction for tombstones.

Because tombstones represent the removal of state, they need to be applied on *all* servers in order to ensure state is properly removed from memory. With the changes in #31, it's not possible for tombstones to be removed from the log prior to being applied on all servers. For instance, imagine the following scenario:
* In a three node cluster, a client submits a `put 1` operation and then a `remove 1` operation. `put 1` is a normal command, and `remove 1` is a tombstone
* The leader commits both operations, but one follower stores only `put 1` before crashing
* The leader and remaining follower compact both operations from their logs
* When the down follower recovers, its log contains `put 1` but the leader and other follower have compacted both `put 1` and `remove 1` from their logs. Thus, the recovered follower will never receive the `remove 1` operation and will therefore maintain an inconsistent state until/if some other operation overrides that state

Essentially, servers need to keep track of separate indexes for safely performing minor and major compaction. Minor compaction removes normal commands from the log and should be based on client event indexes, and major compaction removes tombstones from the log and should be based on the highest index stored on all servers.

The simple solution to this is to reinstate the `globalIndex` tracking solely for the purpose of managing compaction of tombstones. `globalIndex` basically keeps track of the highest index stored on *all* servers. This is significantly less of an issue than the general use of `globalIndex` was previously since tombstones are infrequently compacted from the log anyways.

There may be some alternative approaches to handling this problem that allow tombstones to be removed from the log even while a server is down. One option is to have servers infrequently commit an entry (via the leader) that effectively persists the server's `commitIndex` in the Raft log. Then, the highest committed `commitIndex` for all servers could be used to compact logs while a server is down. This has the benefit of persisting through leader changes, which isn't the case with tracking `globalIndex` in memory on the leader.

kuujo:
This PR fixes #38 by recovering the concept of a `globalIndex`. The `globalIndex` is indicative of the highest index stored on all servers. The global index is used to determine when it's safe to remove tombstones from the log during major compaction. If a server is partitioned, failure to replicate a tombstone entry to that server before compaction will result in inconsistent state across the cluster.

This PR also fixes some issues with the previous implementation of `globalIndex`. Previously, only active cluster members were considered in the calculation of the global index. But because passive members maintain state machines while they catch up to the rest of the cluster, it's important that they be included in counts. Similarly, this PR fixes an issue wherein state for passive servers wasn't reset when a leader first transitioned.

Note that this PR does still allow major compaction to operate normally aside from removing tombstones. It stores two compaction indexes in the log: `minorIndex` and `majorIndex`. The `minorIndex` is the highest index for which normal entries can be safely removed from the log (based on events received by clients), and `majorIndex` is the highest index for which tombstones can be safely removed from the log. If the major compaction process is compacting a section of the log with indexes greater than the `majorIndex`, it will simply retain any tombstones.

I'm going very carefully through the entire project to document and verify the algorithms. This is a big step in the right direction. Excited about finding some good potential issues.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/request/AppendRequest.java
+++ b/server/src/main/java/io/atomix/copycat/server/request/AppendRequest.java
@@ -60,7 +60,8 @@ public class AppendRequest extends AbstractRequest<AppendRequest> {
   private long logIndex;
   private long logTerm;
   private List<Entry> entries = new ArrayList<>(128);
-  private long commitIndex;
+  private long commitIndex = -1;
+  private long globalIndex = -1;
 
   /**
    * Returns the requesting node's current term.
@@ -116,13 +117,23 @@ public class AppendRequest extends AbstractRequest<AppendRequest> {
     return commitIndex;
   }
 
+  /**
+   * Returns the leader's global index.
+   *
+   * @return The leader global index.
+   */
+  public long globalIndex() {
+    return globalIndex;
+  }
+
   @Override
   public void writeObject(BufferOutput<?> buffer, Serializer serializer) {
     buffer.writeLong(term)
       .writeInt(leader)
       .writeLong(logIndex)
       .writeLong(logTerm)
-      .writeLong(commitIndex);
+      .writeLong(commitIndex)
+      .writeLong(globalIndex);
 
     buffer.writeInt(entries.size());
     for (Entry entry : entries) {
@@ -138,6 +149,7 @@ public class AppendRequest extends AbstractRequest<AppendRequest> {
     logIndex = buffer.readLong();
     logTerm = buffer.readLong();
     commitIndex = buffer.readLong();
+    globalIndex = buffer.readLong();
 
     int numEntries = buffer.readInt();
     entries.clear();
@@ -151,7 +163,7 @@ public class AppendRequest extends AbstractRequest<AppendRequest> {
 
   @Override
   public int hashCode() {
-    return Objects.hash(getClass(), term, leader, logIndex, logTerm, entries, commitIndex);
+    return Objects.hash(getClass(), term, leader, logIndex, logTerm, entries, commitIndex, globalIndex);
   }
 
   @Override
@@ -163,14 +175,15 @@ public class AppendRequest extends AbstractRequest<AppendRequest> {
         && request.logIndex == logIndex
         && request.logTerm == logTerm
         && request.entries.equals(entries)
-        && request.commitIndex == commitIndex;
+        && request.commitIndex == commitIndex
+        && request.globalIndex == globalIndex;
     }
     return false;
   }
 
   @Override
   public String toString() {
-    return String.format("%s[term=%d, leader=%s, logIndex=%d, logTerm=%d, entries=[%d], commitIndex=%d]", getClass().getSimpleName(), term, leader, logIndex, logTerm, entries.size(), commitIndex);
+    return String.format("%s[term=%d, leader=%s, logIndex=%d, logTerm=%d, entries=[%d], commitIndex=%d, globalIndex=%d]", getClass().getSimpleName(), term, leader, logIndex, logTerm, entries.size(), commitIndex, globalIndex);
   }
 
   /**
@@ -278,6 +291,18 @@ public class AppendRequest extends AbstractRequest<AppendRequest> {
     }
 
     /**
+     * Sets the request global index.
+     *
+     * @param index The request global index.
+     * @return The append request builder.
+     * @throws IllegalArgumentException if index is not positive
+     */
+    public Builder withGlobalIndex(long index) {
+      request.globalIndex = Assert.argNot(index, index < 0, "global index must not be negative");
+      return this;
+    }
+
+    /**
      * @throws IllegalStateException if the term, log term, log index, commit index, or global index are not positive, or 
      * if entries is null 
      */
@@ -289,6 +314,7 @@ public class AppendRequest extends AbstractRequest<AppendRequest> {
       Assert.stateNot(request.logTerm < 0, "log term must not be negative");
       Assert.stateNot(request.entries == null, "entries cannot be null");
       Assert.stateNot(request.commitIndex < 0, "commit index must not be negative");
+      Assert.stateNot(request.globalIndex < 0, "global index must not be negative");
       return request;
     }
 
--- a/server/src/main/java/io/atomix/copycat/server/response/ConfigurationResponse.java
+++ b/server/src/main/java/io/atomix/copycat/server/response/ConfigurationResponse.java
@@ -72,7 +72,10 @@ public class ConfigurationResponse<T extends ConfigurationResponse<T>> extends A
       activeMembers = serializer.readObject(buffer);
       passiveMembers = serializer.readObject(buffer);
     } else {
-      error = RaftError.forId(buffer.readByte());
+      int errorCode = buffer.readByte();
+      if (errorCode != 0) {
+        error = RaftError.forId(errorCode);
+      }
     }
   }
 
@@ -84,7 +87,7 @@ public class ConfigurationResponse<T extends ConfigurationResponse<T>> extends A
       serializer.writeObject(activeMembers, buffer);
       serializer.writeObject(passiveMembers, buffer);
     } else {
-      buffer.writeByte(error.id());
+      buffer.writeByte(error != null ? error.id() : 0);
     }
   }
 
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -1013,6 +1013,7 @@ final class LeaderState extends ActiveState {
       // If the active members list is empty (a configuration change occurred between an append request/response)
       // ensure all commit futures are completed and cleared.
       if (members.isEmpty()) {
+        context.setCommitIndex(context.getLog().lastIndex());
         for (Map.Entry<Long, CompletableFuture<Long>> entry : commitFutures.entrySet()) {
           entry.getValue().complete(entry.getKey());
         }
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -233,6 +233,8 @@ final class LeaderState extends ActiveState {
       index = context.getLog().append(entry);
       LOGGER.debug("{} - Appended {} to log at index {}", context.getAddress(), entry, index);
 
+      // Store the index of the configuration entry in order to prevent other configurations from
+      // being logged and committed concurrently. This is an important safety property of Raft.
       configuring = index;
       context.getCluster().configure(entry.getIndex(), entry.getActive(), entry.getPassive());
     }
@@ -241,6 +243,7 @@ final class LeaderState extends ActiveState {
     replicator.commit(index).whenComplete((commitIndex, commitError) -> {
       context.checkThread();
       if (isOpen()) {
+        // Reset the configuration index to allow new configuration changes to be committed.
         configuring = 0;
         if (commitError == null) {
           future.complete(logResponse(JoinResponse.builder()
@@ -306,6 +309,8 @@ final class LeaderState extends ActiveState {
       index = context.getLog().append(entry);
       LOGGER.debug("{} - Appended {} to log at index {}", context.getAddress(), entry, index);
 
+      // Store the index of the configuration entry in order to prevent other configurations from
+      // being logged and committed concurrently. This is an important safety property of Raft.
       configuring = index;
       context.getCluster().configure(entry.getIndex(), entry.getActive(), entry.getPassive());
     }
@@ -314,6 +319,7 @@ final class LeaderState extends ActiveState {
     replicator.commit(index).whenComplete((commitIndex, commitError) -> {
       context.checkThread();
       if (isOpen()) {
+        // Reset the configuration index to allow new configuration changes to be committed.
         configuring = 0;
         if (commitError == null) {
           future.complete(logResponse(LeaveResponse.builder()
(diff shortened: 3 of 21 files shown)


kuujo:
This seems to be pretty close, but I'm going to go through it one more time and try to break it a bit more before it can be merged.

kuujo:
Same test hanging that's been hanging randomly for a while. I'll have to look into that.

kuujo:
Gonna try to fix the tests for all these PRs before they're merged, but they can be reviewed.

kuujo:
So, it seems Travis is killing the Maven tests due to memory consumption:
https://github.com/travis-ci/travis-ci/issues/2109

I had to reduce some of the load in the tests to resolve this issue. I suspect the above tests will pass now.

kuujo:
@madjam can you review this? Does this make sense?

I'm in the process of documenting session events and log compaction. I think this and #31 are a huge improvement and I haven't found any other potential issues yet.

kuujo:
@jhalterman is almost done with configuration changes in Jepsen tests. Once all these PRs are merged we can do those tests.

madjam:
@kuujo This is very nice. I will be looking into this today. Finally wrapped up some other work I was doing.

madjam:
So if I understand this right the major compactor treats log entries at index greater than `majorIndex` just like a minor compactor would treat all its log entries. But otherwise it (the major compactor) can safely remove tombstones because those entries are all globally replicated?

I'm trying to better understand how the following scenario would work:
- `3` node cluster with `majorIndex` at `100`
- `4`th node joins.
- Leader is replicating log entries to the new node (in `PASSIVE` state) while its major compactor is running

Is there a possibility for inconsistency to arise here?

kuujo:
@madjam So, there is definitely a potential issue here. Basically, state up to the `globalIndex` cannot be considered to be consistent across servers. Logs will always result in the same state once applied up to the commit index, but log cleaning and differences in the logs across servers means arbitrarily composing them can result in an inconsistent state as far as I can tell. That effectively means if a leadership change occurs when a server is in `PASSIVE` state and before the `PASSIVE` server's log contains entries up to `globalIndex`, the passive server's log can become inconsistent because compaction on different servers can happen at different times. The first leader may send an entry that the second server already cleaned and for which a tombstone was already removed.

So, I think one simple and obvious solution to this is to simply place a constraint on `PASSIVE` servers that ensures they're only ever caught up by one leader. Essentially, servers in the `PASSIVE` state would simply keep track of the `AppendRequest.leader()` in memory for the first entry in their log. If they receive an `AppendRequest` from a different leader containing entries less than `globalIndex`, the truncate their log and the leader replays all entries from the start. Once a `PASSIVE` member's log is caught up to `globalIndex` it's safe for leadership changes to occur.

This is a simplistic but undoubtedly correct solution. Obviously, the downside is that leadership changes are handled rather poorly, but that may be acceptable for catching up joining servers.

I *think* joining servers pose a consistency issue, but I'm going to try to come up with some concrete scenarios (more extensive than above) to demonstrate this issue this weekend. The are a lot of variables that haven't been considered that may or may not contribute to consistency issues here. There are also potentially some more complex but more elegant solutions that don't cause the leadership change issue above with requiring that a `PASSIVE` server be caught up by a single leader. For instance, one option is to make log compaction deterministic by having the leader commit a `CompactEntry` or something like that. But in my mind, this may be less ideal than the above because it prevents some I/O from happening independently on each server and would require that configurations be replicated too to ensure that each server compacts the same entries from its logs. I imagine I'll come up with some more options this weekend.

madjam:
I like your idea of placing some restrictions on how `PASSIVE` members are caught up to the rest of cluster. Even though it is not the most optimal in terms of data transfer, I think it is safe.

Another possibility I was thinking about was for all `ACTIVE` members to loosely coordinate their major compactions. This coordination will need to ensure that that the `majorIndex` on every `ACTIVE` member is less than or equal to the current leader's `majorIndex`. So if a leadership change occurs whilst a PASSIVE member is building up its log, we can provide a guarantee that the next leader would have not evicted entries that need to be sent to the `PASSIVE` member.

kuujo:
I think you're thinking in the right direction. At least we know the leader restriction, which not ideal, is certainly safe and may be a good option initially. But I'm going to try to put together a good safety argument for an approach to allowing independent compaction and handling leader changes when catching up a server tonight. I actually think it's somewhat questionable whether there is actually a consistency issue here because of the semantics of how entries are cleaned and logs are compacted. For instance, in the scenario above, I think it may be possible to modify the way the log exposes entries such that the first server *can't* send an entry that's followed by a tombstone. We have the distinct advantage of knowing the leader's log is more up to date than its followers, and that I think can be used to prevent such cases. I'm going to try to prove that and if it's too complex or feels like some scenarios may still be unaccounted for I'll implement the leader restriction I mentioned in the interest of safety in the near term.

Will post back here likely later tonight.

kuujo:
@madjam actually, I think the leader restriction is rather acceptable if you consider the following: in standard Raft implementations that take snapshots, when a server joins the cluster, the leader must send an initial snapshot to the joining server before replaying logs. Presumably, if the a leadership change occurs, the joining server must fail or the new leader starts sending its own snapshot as snapshots on different servers are taken independent of each other.

In this case, entries up to `globalIndex` effectively represent a snapshot of the leader's state machine. So, I think it's practical to place this restriction on `PASSIVE` servers, and this can certainly be revisited in the future if a better solution arises.

There is some potential to optimize this algorithm in the future as well. Currently, even after entries are `clean()`ed from the log by state machines, they're still made visible to the Raft algorithm in order to ensure tombstones are replicated and state is cleared. But once an entry has been stored on all servers and is cleaned, it could be immediately hidden so that it doesn't have to be unnecessarily sent to `PASSIVE` servers. But there's a consistency issue here that is the same consistency issue addressed in the `MajorCompactionTask`. I think that documentation explains the race condition pretty well, but essentially it's possible that as it iterates through the log the leader could send an active entry that's later cleaned by a tombstone, and if the tombstone isn't also sent (because it's less than `globalIndex`) we would still get an inconsistency. So, this would require that a snapshot of the state of entries in the log be taken before sending entries to a `PASSIVE` server in order to ensure they don't change while the log is being iterated. But because the major compaction process handles this internally in order to ensure consistency, we can rely on the compaction process to maintain consistency in the log rather than in the leader. I think it would be overly complex to try to do the same thing for catching up `PASSIVE` servers at least in the short term, and presumably tombstones are a smaller portion of the log anyways (normal commands can still be safely removed by minor and major compaction).

kuujo:
I'm going to update this PR with the leader restriction tonight

kuujo:
To map this out, the following graphic shows the inconsistency that can happen from logs being compacted on different servers at different times:

inconsistency

In the illustration, entries that are grey have been `clean()`ed by the state machine but not removed in log compaction, so they're available to be sent to any other server if necessary. Entries that are yellow are entries that have been applied to and are still active in the state machine. Entries that are white with a solid border have been compacted from the log. Indexes that are white with a dotted border are not present in the log. When the state machine encounters a normal value, it cleans any prior instance of that identifier. When the state machine encounters a `nil` value, it cleans the `nil` and all prior instances of that identifier.

Essentially, this depicts a three node cluster with `server 1`, `server 2`, and `server 3`. Servers `1` and `2` each have `10` entries in their log. Server `2` compacted its logs up to index `7`, but none of the other servers have compacted their logs. A fourth server, `server 4`, joins the cluster and begins receiving entries from `server 1`. Server `1` crashes and server `2` is elected and continues sending entries from where server `1` left off. The inconsistency occurs when server `1` sent entry `6` with the value `a←1` though it was actually removed by a tombstone later in the log, and server `2` has already compacted the tombstone from its logs and thus never sends it to server `4`.

However, if we modify the logic to immediately remove all `clean()`ed entries up to the `globalIndex` (`7` in this case) we resolve these types of issues. Essentially, all grey indexes in the above illustration would become white up to index `7` and would not be sent to the `PASSIVE` server. But even in this case, I wondered whether a condition could occur while iterating through the log that would result in the leader sending server `4` `a←1` without the later related tombstone `a←nil` if that tombstone (and the prior command) is cleaned after the command is sent. The answer I came up with is: no. The reason is because of the semantics of log cleaning. The log cleaning process assumes that for each command, prior related commands will be immediately `clean()`ed. So, if an `a←1` and then `a←nil` command are in the log, when the `a←nil` command is applied, `a←1` will be cleaned even if `a←nil` is not immediately cleaned.

The essential invariants here are:
* The `globalIndex` will never increase while a `PASSIVE` server is alive and has not received entries up to the `globalIndex`
* Tombstones will never be removed or hidden for indices `> globalIndex`

So, given these constraints, I haven't yet been able to produce any scenario where server `1` sends `a←1` to server `4` and server `2` fails to send `a←nil` after a leadership change if entries are immediately hidden upon being `clean()`ed.

Scenario 1:
* `a := set("foo")` at index `2`
* `b := delete("foo")` at index `100`
* `globalIndex` is `101` at the start of the join
* `a` will be skipped since it was cleaned when `b` was applied
* `b` will be skipped since it's a tombstone with an index `< globalIndex`

Scenario 2:
* `a := set("foo")` at index `2`
* `b := delete("foo")` at index `100`
* `globalIndex` is `99` at the start of the join
* `a` will be skipped since it was cleaned when `b` was applied
* `b` will be sent since it's a tombstone with an index `> globalIndex`

Scenario 3:
* `a := set("foo")` at index `2`
* `globalIndex` is `99` at the start of the join
* `b := delete("foo")` is committed and applied at index `100` during the join and cleans `a`
* `a` may or may not be sent to the `PASSIVE` server
* `b` is guaranteed to be retained and sent to the `PASSIVE` server since `globalIndex` is less than `b`s index and cannot increase until the `PASSIVE` server is caught up

I believe these scenarios to prove that because of the semantics of log cleaning in Copycat, it is safe to catch joining servers up normally - even with leadership changes - if all `clean()`ed entries with indexes less than the `globalIndex` are not replicated even if they're still stored on disk. However, because of the consistency issue depicted in the illustration above, I think `globalIndex` will have to be persisted in a periodic log entry in order to ensure it never goes back in time. In the event that a cluster is restarted *after* compaction, the above inconsistency could occur if a server joins the cluster before the leader is able to resolve the accurate `globalIndex` (which it wouldn't be able to do if a server were down). Essentially, this means `globalIndex` should only be increased by the leader.

kuujo:
TL;DR I know. This is probably the last big algorithm decision to be made though :-) We can make the other PRs short and sweet.

madjam:
@kuujo This looks good to me. Couldn't find anything wrong with your reasoning here.

This does result in `ACTIVE` members unable to advance their `globalIndex` as long as there is at least one `PASSIVE` member in catch up mode. So it is possible for a `PASSIVE` member that joins and then crashes to disrupt the progress of log cleaning on all `ACTIVE` members. May be as a further optimization the implementation might forcibly evict `PASSIVE` members that went AWOL. When they rejoin later they need to first truncate all their log entries and start afresh. I feel this optimization if it makes sense can be added later. But your core algorithm looks safe to me.

kuujo:
Good point about a `PASSIVE` member disrupting `globalIndex`.

So, one thing that hasn't been implemented in Copycat but is recommended in the Raft dissertation is limiting the amount of time a joining server can take to be caught up. The idea is to handle cases where the cluster is operating faster than the joining server such that it could never be caught up. But perhaps the same or a similar strategy could be adopted to make sure `globalIndex` continues to advance.

Diego suggests an algorithm wherein the leader caps the number of AppendEntries RPCs that can be used to catch up the new server. If at the end of *n* rounds the server isn't caught up to within a heartbeat of the leader, the join operation essentially fails and the new server has to try again.

Really, though, the idea of when/how to determine if a joining server can't be caught up is sort of an area that doesn't have much potential for safety concerns and doesn't have a fixed algorithm, only suggestions. I suppose it depends on the intended behavior of the system. Do we want to ensure that a join is only cancelled if the server *can't* be caught up, or if it can't be caught up within some fixed time frame? If the later, what is that fixed time frame? Really, in terms of determining whether a joining server can and will be caught up to the leader, what I really think needs to be determined is just whether the `PASSIVE` server's last log index is advancing such that it will eventually converge on `commitIndex`. That is, for some set of AppendEntries RPCs within some time frame, the leader's `commitIndex` minus the `PASSIVE` server's last log index has decreased. But I digress. This particular aspect of configuration changes should be left to another issue and PR.

But I do think it's prudent to ensure a joining server can't prevent `globalIndex` from advancing. Presumably, even if some mechanism for limiting the amount of time a server can take to be caught up (as is suggested in the Raft dissertation) is *not* implemented right now, that still won't prevent `globalIndex` from advancing since a `PASSIVE` server will eventually be caught up to the fixed `globalIndex` as long as we ensure that it is indeed advancing at some rate.

So, absent the Raft join timeout, I suggest we simply add logic to the leader that ensures that each `PASSIVE` server's `matchIndex` advances within a given election timeout. That is, if a `PASSIVE` server can't be reached (is partitioned or crashed) for an election timeout, it's removed from the configuration to allow `globalIndex` to advance.

Some mechanism will also have to be put in place to ensure a `PASSIVE` member that is partitioned and gets removed from the configuration also fails. For instance, perhaps a `PASSIVE` server could use the election timeout in the same way that candidates use it. If it hasn't received an `AppendRequest` in an election timeout, it attempts to rejoin the cluster. If it was removed from the configuration, as you mentioned, when it's added again a second time the leader will force it to truncate its log again.

I'm not sure I'll do all that now. We'll see how this initial modification goes. We finished writing Jepsen tests for configuration changes, so will run those against the modifications as well this weekend.

kuujo:
@madjam I think the above should suffice.

kuujo:
Cool thanks! I'm gonna run some Jepsen tests with configuration changes and then merge it maybe tomorrow.

PR #49, opened 2015-10-29
Added a new testing module to house testing infrastructure.

jhalterman:
Made test module not build a test jar, since nothing else is using it.
Moved testing related dependencies from parent pom to testing module.


jhalterman:
If we want testing to release with the other modules, it needs to inherit from the parent pom, which will create a circular dependency with the current setup. Closing for now.

PR #52, opened 2015-11-02
Logging improvements.

madjam:
Couple of minor logging updates:
    - Provide more context in the log message when AE request to a member fails.
    - Lower log level in compactor to avoid excessive logging at info level


kuujo:
:+1:

PR #53, opened 2015-11-03
Move persistent and volatile server state into the log

kuujo:
This PR moves persistent and volatile server state into the log. The reasoning behind this is because the Raft algorithm specifies that `term` and `votedFor` should both be written to disk. So, we add a `LogDescriptor` to the log and write term and election information to that file. In order to ensure the API isn't spread around too much, I decided to simply move all the state for a server into the `Log`. This doesn't totally feel right, but there's bound to be some awkwardness since only *some* server state is persisted and some is held in memory only.

I don't intend to merge this pull request immediately, but I want to start some discussion on the design since it feels a bit dirty, but maybe that's all in my head. The new `Log` methods need more tests, and tests for recovering `term` and `vote` should be added to recovery tests as well.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/CopycatServer.java
+++ b/server/src/main/java/io/atomix/copycat/server/CopycatServer.java
@@ -178,7 +178,7 @@ public class CopycatServer implements RaftServer {
 
   @Override
   public long term() {
-    return state.getTerm();
+    return state.getLog().getTerm();
   }
 
   @Override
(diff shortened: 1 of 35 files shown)


madjam:
My first reaction is that this info does not belong to the Log's domain. My justification it simply based on idea that Log in its entirety is a unit of replication. The protocol's goal is to keep the Log identical on all its members and therefore it might be cleaner track member specific non-volatile state separately outside of the Log.

Do you see challenges with tracking this state outside?

kuujo:
I definitely agree. I dislike it. I think the log should store only individual mutations and some other mechanism should store server state. It doesn't feel right that volatile server state is stored here.

So, i can scrap this branch and revert to the previous log. In that case, rather than adding server state methods to the log (term, leader, vote) maybe we should create A separate class for storing server state that's also managed by `Storage` but is not the `Log`. So, the `Log` and sever metastore will be separate, but both can still be managed by `Storage` to use the same directory/`StorageLevel` configurations. That actually feels a lot nicer to me.

PR #54, opened 2015-11-17
Truncate segments in reverse order to prevent gaps in the log

kuujo:
This PR modifies the `Log.truncate` method to truncate segments in reverse order. This is necessary in order to prevent gaps in the log if a failure occurs while truncating the log. If multiple segments are truncated from the log starting at the first truncated segment, a gap can exists, and in that case the log will interpret the gap as compacted entries. So, this ensures that only segments from the end of the log will have been removed in the event of a failure during `truncate`.


madjam:
:+1:

PR #55, opened 2015-11-17
Make no-op entries tombstones

kuujo:
This PR marks `NoOpEntry` as a tombstone. This is necessary to ensure that no-op entries are stored and applied on *all* servers. The reason that's critical is because no-op entries now reset session timeouts at the beginning of the leader's term. If a no-op entry is cleaned from the log and not sent to some follower, that follower's session state can become inconsistent if a session expires immediately after the no-op entry.


madjam:
+1

PR #56, opened 2015-11-18
Set globalIndex to one less than the highest log entry replicated to …

madjam:
…all cluster members

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -1005,7 +1005,17 @@ final class LeaderState extends ActiveState {
       // passive members. This is critical since passive members still have state machines and thus it's still
       // important to ensure that tombstones are applied to their state machines.
       // If the members list is empty, use the local server's last log index as the global index.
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
```diff
@@ -1005,7 +1005,17 @@ private void commitEntries() {
...
+
+      // The globalIndex should always be one less than the highest index that is replicated to all members.
+      // This is to ensure prevEntry checks can always be performed for AppendRequests.
+      // Consider the following scenario:
+      // - Entry at index i is a NoOpEntry and it is fully replicated to all members. (NoOpEntries can be garbage collected
+      // as soon as they are replicated to all cluster members.)
+      // - A leadership change occurs and the new leader sends an AppendRequest with logIndex == i.
+      // Any follower should be able to validate that they have this entry in their log with a term that matches
+      // what the new leader is advertising.
+      context.setGlobalIndex(Math.min(globalMatchIndex - 1, 0));
```
kuujo: This should be `max` right? If `matchIndex` for any server is `0` (it has received no entries) then `globalMatchIndex` will be `0` and so we'll get `-1` and that may fail validation somewhere.

madjam: Oops. You are right.

kuujo:
:+1:

Issue #57, opened 2015-11-18
KeepAlives can stall after a connection reset


madjam:
I have observed this several times running with a 3 node cluster set up. The trigger is usually a request failure (usually due to a timeout) that causes the connection to be reset and from that point on no more KeepAliveRequests are sent. This eventually leads to the session expiring.

My suspicion is that there is an uncaught exception that is causing this. By changing how new connections are established after a reset I was able to "fix" this problem. I'll clean up that change and submit a PR shortly.

PR #58, opened 2015-11-18
Fix race condition in client connection reestablishment.

madjam:
Fixes https://github.com/atomix/copycat/issues/57


kuujo:
Nice. Seems good to me!

PR #61, opened 2015-11-19
Ensure one committed entry is always accessible in the log

### LINKED ISSUE #60: Handle compacted entries during AppendRequest consistency checks
i think the fix for handling `null` (cleaned) entries during `AppendRequest` consistency checks may be wrongheaded:
500186887ba3f8001e6d387b9ea70b920109dfcb

So, to redefine the problem, when checking the term of the previous entry while handling an `AppendEntries` RPC, if the entry has been cleaned it will be `null`. That causes the consistency check to fail. What we did in the fix for this way ensure that `globalIndex` is always the lowest `matchIndex` minus `1`, but this only fixes the problem for tombstones. Normal entries can still be `clean()`ed and thus hidden from reading after the `globalIndex`. So it's likely this problem would still occur if the `NoOpEntry` in our scenario were not changed to a tombstone.

I think the case is actually that we *could* safely hide all committed and cleaned entries. If an entry returns `null` from the log, we know it must necessarily have been committed and therefore consistency checks are not necessarily even needed. Any leader that's elected will have the same entry at the same term simply as a product of the Raft election algorithm.

But I think just to feel good about it we should still keep the last committed entry in the log visible at all times. As I mentioned, the log compaction process already ensures that the last committed entry always remains on disk and will never be removed, so I think I was wrong and you were right @madjam that we should simply ensure the last committed entry in the log always returns a non-null entry. So, a property of the log would be that entries from `commitIndex` to `lastIndex` are always non-null, and that would make this feel a lot better.

kuujo:
This PR fixes #60 to ensure that at least one committed entry is always retained in the log and made accessible during `AppendRequest` handling.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/LeaderState.java
@@ -1006,16 +1006,7 @@ final class LeaderState extends ActiveState {
       // important to ensure that tombstones are applied to their state machines.
       // If the members list is empty, use the local server's last log index as the global index.
       long globalMatchIndex = context.getCluster().getMembers().stream().mapToLong(MemberState::getMatchIndex).min().orElse(context.getLog().lastIndex());
-
-      // The globalIndex should always be one less than the highest index that is replicated to all members.
-      // This is to ensure prevEntry checks can always be performed for AppendRequests.
-      // Consider the following scenario:
-      // - Entry at index i is a NoOpEntry and it is fully replicated to all members. (NoOpEntries can be garbage collected
-      // as soon as they are replicated to all cluster members.)
-      // - A leadership change occurs and the new leader sends an AppendRequest with logIndex == i.
-      // Any follower should be able to validate that they have this entry in their log with a term that matches
-      // what the new leader is advertising.
-      context.setGlobalIndex(Math.max(globalMatchIndex - 1, 0));
+      context.setGlobalIndex(globalMatchIndex);
 
       // Sort the list of replicas, order by the last index that was replicated to the replica. This will allow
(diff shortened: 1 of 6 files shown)


kuujo:
I think the first build failed due to the Catalyst bug recently fixed. I released a new Catalyst version, so I'll update the dependency in another PR.

madjam:
:+1:
Nice. I like this much better.

kuujo:
Agree. Much nicer.

But I think there could still be an issue here. `commitIndex` can be greater than `lastIndex` and in that case, this logic will hide `lastIndex` if it's `clean()`ed. I'm actually not sure if this could happen in practice. State machines should clean entries deterministically, and `AppendRequest` doesn't send `null` entries but always ends with a non-null entry. So, it shouldn't really be possible for a log with `commitIndex > lastIndex` to have a non-null entry at the end since the only way for an entry at the end of a log to be null is for the *local* server state machine to `clean()` it, and the local server state machine can't clean an entry before the leader's state machine simply as a property of the order of commitment. But maybe it wouldn't hurt to add a `index == lastIndex()` condition just in case.

madjam:
I agree. This is a case that should not arise. It may make sense to add an assertion to catch it if/when it occurs.

kuujo:
Ahh yes assertion good idea

kuujo:
Hmm... The assertion's kind of a pain because some code *does* read the log in that state while appending entries. I'm just gonna leave it how it is for now.

On a related note, the fact that it writes entries sequentially and entries received in `AppendRequest` can be received with gaps does not impact the assertion that the last entry will always be visible. If a failure occurs while handling an `AppendRequest`, the recovery process will recover the log up to the last written non-null entry and will exclude skipped/null entries, so that's not an issue.

Issue #51, opened 2015-11-01
Prevent split brain on startup


kuujo:
The current implementation of configuration changes handles startup of the cluster rather poorly. Essentially, when each server is started, it attempts to find an existing leader to join the cluster. If it can't find a cluster to join, it assumes it's a member of a new cluster and transitions directly to follower.

This can allow split brain when some number of joining servers can't see an existing leader. For instance, if a single node is running in the cluster and two new servers are added, if those two servers can't see the existing server (a single leader), they'll form their own cluster and elect a leader. Neither side of the partition will know about the other, and both sides will accept writes.

I propose that the structure of server configurations be used to determine whether a server is expected to already be a member of a cluster or if its joining a cluster. There's no safety issue of a server thinks it should be joining when it's already a member of a cluster, but there is a safety issue if a server thinks it's starting a new cluster when one already exists.

If the server's `Address` is listed in the cluster configuration, it should transition directly to `FOLLOWER` and assume that it's already a member of the cluster. If the follower does not receive any heartbeats from a leader within an election timeout and transitions to candidate and requests votes from other servers, those servers should verify that the candidate is present in their configurations. If a server is not present in some other server's configuration, it should transition back to `INACTIVE` and fail since this may indicate that it should have joined the cluster but was instead started as a full member.

Raft suggests servers should form a cluster by joining one by one. Initially, a single server cluster starts, and then additional servers are added through configuration changes. From an API perspective, something like this could perhaps be done by exposing methods that differentiate between joining and leaving a cluster of starting the server as an existing member of the cluster. As long as safety mechanisms are in place this would be fine.

kuujo:
Actually, the addition of separate client/server `Address`es may necessitate single member configuration changes only. It's likely that with separate client/server addresses, each server will need to be configured with a list of *server* addresses and a local client address. In that case, client addresses for all servers must be shared via configuration changes so that clients can learn about new servers in `KeepAliveResponse`s.

kuujo:
I have a big PR on the way for this and a couple other issues related to handling configuration changes

kuujo:
This is fixed in #62

PR #62, opened 2015-11-21
Support dynamic replacement of voting members

kuujo:
This PR is the last significant refactoring that will be done before the 1.0 release. Once this is merged, a release candidate will be released.

The idea here is to allow clusters to dynamically resize themselves according to a configured `quorumHint`. With this change, servers can be in one of three states:
* `ACTIVE` - active voting members of the Raft cluster. The cluster will always attempt to maintain `quorumHint` active members
* `PASSIVE` - passive members are members that receive `AppendRequest`s from followers and maintain a state machine. In the event that an `ACTIVE` server becomes unavailable, it can be quickly replaced by the most up-to-date `PASSIVE` member
* `RESERVE` - reserve members are stateless members that simply await promotion. Reserve servers receive only configuration updates from followers

This results in a cluster that can rapidly replace unavailable Raft members with available passive members. Each server periodically sends a `HeartbeatRequest` to the leader, and heartbeats are tracked in a manner similar to sessions. When an `ACTIVE` member stops sending heartbeats to the leader, it will, if possible, be demoted to `RESERVE` and a `PASSIVE` or `RESERVE` server will be promoted to `ACTIVE` to take its place. The promotion will happen by *first* promoting a new `ACTIVE` member and then demoting the `UNAVAILABLE` voting member.

This, in theory, allows large clusters to be started and operated with little management. Servers can join and leave the cluster at will, and the system will always attempt to maintain exactly `quorumHint` voting members. This is illustrated in the `testReplace` test method:
https://github.com/atomix/copycat/blob/dynamic-configurations/test/src/test/java/io/atomix/copycat/test/ClusterTest.java#L105-L122

This test creates a 3 node cluster, adds three nodes to the cluster, and then removes the original three nodes, all while submitting writes. Initially, the 3 nodes are added as `RESERVE` and then promoted to `PASSIVE` members (the `backupCount` dictates the number of passive members). As `ACTIVE` servers are shut down, `PASSIVE` members take their place to allow writes to continue without losing fault tolerance. (This is actually not implemented totally correctly at the moment. While unavailable servers are replaced by first promoting and then demoting, when an active server leaves the cluster it is first removed from the configuration before a new member is promoted. This should not be the case).

The use case for this feature is to simplify the management of embedded Copycat clusters. The number of servers in the cluster does not need to relate to the Raft algorithm in any way. Only the `quorumHint` and `backupCount` need relate to the Raft algorithm. One place where this is useful is in [Vert.x](http://vertx.io) (to which I'm a significant contributor). With the additional scalability, an Atomix Vert.x `ClusterManager` becomes much more approachable for Vert.x users who are used to embedded HA systems which scale well.

Note that the use of dynamic quorums is entirely optional. Starting a cluster in the typical way will still yield the expected result. If no `quorumHint` is specified, all configured members are assumed to be Raft voting members. Once the cluster has been started, the quorum hint remains the same size as the initial number of core members, and any additional servers will be added in the reserve/passive state. So, it's possible to scale down the number of voting members, but it cannot currently be scaled up beyond the initial configuration.

Because of the significant refactoring that went into this, this PR also fixes a couple other issues. It adds a `MetaStore` which stores the term, vote, and configuration of each server on disk. Additionally, a significant change in this PR is that servers now have separate `Server`s for client and server communication. So, each server can define both a client and server `Address`. Clients receive client `Address`es through `KeepAliveResponse`s, and servers propagate that information to each other in configurations. I wanted to get the client `Address` feature in before the 1.0 release because a big goal of post-1.0 development will be to expose an HTTP/REST interface for clients. So, ultimately the client interface will provide an abstraction to allow clients to interact with servers using different protocols without forcing servers to use the same protocol.

@madjam I'm assigning this to you. Let me know if anything seems unclear. I'm pretty tired right now and feel like what I'm writing is all over the place. What are your thoughts on this?


kuujo:
Need to update API documentation for this behavior.

kuujo:
One final note on configuration changes. Previously, it was possible for split brain to occur in the event of a partition during startup of more than one node (#51). Servers essentially attempted to look for an existing cluster and if they couldn't find one they would form their own cluster. This can obviously result in servers arbitrarily forming their own clusters, but that is no longer an issue. All servers immediately persist their initial configuration on startup, and all servers use their most recent configuration at startup. If a member is listed in the initial members list, it assumes it's an `ACTIVE` member and starts as such. If a member is not listed in the initial members list, it will expect to join a cluster. Thereafter, its state is defined by the configuration stored on disk.

kuujo:
One more significant change I just noticed: I wasn't too fond of changing the interfaces, but I changed `CopycatClient` and `CopycatServer` to interfaces and `CopycatRaftClient` and `CopycatRaftServer` to implementations. The reasoning behind this is because `RaftClient` and `RaftServer` don't feel proper as Copycat's interfaces since in reality they extend the Raft algorithm with additional states and other behaviors that are really more specific to Copycat.

kuujo:
6d95813 made separate client/server `Address`es optional. If only one `Address` is provided, clients and servers will both connect to that address.

kuujo:
I think I forgot to add a couple tid bits:

In a typical Raft implementation, when a server joins the cluster, the leader sends it append requests to catch it up. But with this PR, *followers* are responsible for sending append requests to passive members. This is done by consistent hashing. Each time the configuration changes, each follower uses an algorithm to recompute the passive and reserve servers to which it should send entries and configurations. This consistent hashing includes only `AVAILABLE` members. Availability is determined by the heartbeats mentioned above.

This reduces the load on the leader and spreads the responsibility for replication across all the `ACTIVE` nodes. But it also poses another problem: when configuration changes occur, only a majority of Raft nodes know about a configuration when it's committed. So, this can result in passive/reserve servers thinking they should be in a different state than the follower. So a follower will try to send entries to a reserve server, for instance. To mitigate this problem, prior to sending entries to a server, followers will always first send their configuration via a `ConfigureRequest`.

kuujo:
One last point. This also allows tombstones to be removed from logs while a server is down. Essentially, the heartbeat mechanism is used to determine availability and demote unavailable servers. When a server is transitioned to `RESERVE`, it must truncate its log and start over once it gets promoted again. This allows `globalIndex` to be calculated only for non-reserve servers.

madjam:
@kuujo Just got back from a break. I will look into this today.

kuujo:
No worries. It's a big PR. Still doing some testing and sprucing up some of the documentation.

kuujo:
Hmm... looks like those tests are failing pretty regularly now. Not sure what changed. I'm able to get `testThreeOfFiveNodeSubmitCommandWithNoneConsistency` to fail fairly frequently, but not while debugging.

kuujo:
I'm able to get the failing test against 22f8f99 so this bug must have been introduced a while back.

kuujo:
Bah, dumb mistake. Fixed in f4695e4

kuujo:
On a related note, with the leader and follower replication logic separated into `LeaderAppender` and `FollowerAppender` respectively, we should be able to more easily write some tests for that critical logic. I'll try throwing something together for that.

client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
```diff
@@ -580,12 +589,30 @@ private ClientSession resetConnection() {
...
   /**
+   * Selects the next member to which to send a request.
+   */
+  private Address selectMember() {
+    // Remove the next member that meets the strategy from the members list.
+    Iterator<Address> iterator = connectMembers.iterator();
+    while (iterator.hasNext()) {
+      Address member = iterator.next();
+      if (connectionStrategy.connect(member, leader)) {
+        iterator.remove();
```
madjam: I feel we should consider the situation where a client can start up before the active members do. By discarding members that have been tried without success we'll eventually exhaust the list of servers and when that happens the client gives up. It would be good to make this configurable so that a client can choose to retry until a valid session can be established.

protocol/src/main/java/io/atomix/copycat/client/request/KeepAliveRequest.java
```diff
@@ -20,11 +20,20 @@
...
 import java.util.Objects;
 
 /**
- * Protocol keep alive request.
+ * Session keep alive request.
+ * <p>
+ * Keep alive requests are sent by clients to servers to maintain a session registered via
+ * a {@link RegisterRequest}. Once a session has been registered, clients are responsible for sending
+ * keep alive requests to the cluster at a rate less than the provided {@link RegisterResponse#timeout()}.
+ * Keep alive requests also server to acknowledge the receipt of responses and events by the client.
```
madjam: s/server/serve

protocol/src/main/java/io/atomix/copycat/client/request/QueryRequest.java
```diff
@@ -26,7 +26,21 @@
...
+ * the replicated state machine. Each query request must be associated with a registered
+ * {@link #session()} and have a unique {@link #sequence()} number within that session. Queries will
+ * be applied in the cluster in the order defined by the provided sequence number. Thus, sequence numbers
+ * should never be skipped. In the event of a failure of a query request, the request should be resent
+ * with the same sequence number. Queries are guaranteed to be applied in sequence order.
+ * <p>
+ * Query requests should always be submitted to the server to which the client is connected. The provided
+ * query's {@link Query#consistency() consistency level} will be used to determine how the query should be
+ * handled. If the query is received by a follower, it may be evaluated on that node if the consistency level
+ * if {@link io.atomix.copycat.client.Query.ConsistencyLevel#CAUSAL} or {@link io.atomix.copycat.client.Query.ConsistencyLevel#SEQUENTIAL},
```
madjam: s/if/is

client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
kuujo: Actually, it should do that. Effectively, what happens is the client creates the `connectMembers` list. It first tries to connect to members that are allowed by the strategy, and then attempts to connect to members not allowed by the strategy. This ensures that e.g. the `FOLLOWER` strategy can connect to the leader of the cluster is resized down to one node.

So, the `connectMembers` list is used essentially for one iteration of attempts. Once allbers have been attempted (the members list is empty) it will schedule a retry. This is from `ClientSession` (I'm on my phone right now and having trouble linking :-)

```java
    // If we've run out of servers to which to attempt to connect, determine whether we should expire the
    // session based on the responses from servers with which we did successfully communicate and the
    // time we were last able to successfully communicate with a correct server process. The failureTime
    // indicates the first time we received a NO_LEADER_ERROR from a server.
    if (connectMembers.isEmpty()) {
      // If open checks are not being performed, don't retry connecting to the servers. Simply fail.
      if (!checkOpen) {
        LOGGER.warn("Failed to connect to cluster");
        future.completeExceptionally(new IllegalStateException("session not open"));
      }
      // If retries have already been scheduled, queue a callback to be called to retry the request.
      else if (retryFuture != null) {
        retries.add(() -> {
          LOGGER.debug("Retrying: {}", request);
          request(request, future, true, true);
        });
      }
      // If all servers indicated that no leader could be found and a session timeout has elapsed, time out the session.
      else if (failureTime > 0 && failureTime + timeout < System.currentTimeMillis()) {
        LOGGER.warn("Lost session");
        resetConnection().onExpire();
        future.completeExceptionally(new IllegalStateException("session expired"));
      }
      // If not all servers responded with a NO_LEADER_EXCEPTION or less than a session timeout has expired,
      // schedule a retry to attempt to connect to servers again.
      else {
        LOGGER.warn("Failed to communicate with cluster. Retrying");
        retryFuture = context.schedule(Duration.ofMillis(200), this::retryRequests);
        retries.add(() -> request(request, future, true, true));
      }
      return future;
    }
```
So, the logic in the event that all members have been tried is essentially this:
* If no session is open (`checkOpen` is false) simply fail
* If a session is open and less than a session timeout has elapsed, schedule a retry. This ensures clients don't pound servers during leadership changes, but it also ensures that clients sessions will expire when they can't reach the cluster. This is necessary to ensure that e.g. a client that thinks it has a lock will release it even if it can't reach the cluster (which may give the lock to another client).

But first, back to `connectMembers`. For each retry, `connectMembers` is reset. So, that means clients will repeatedly retry all the servers until they believe their session to be expired. See `resetMembers` which resets the members list to a full members list any time a request is successful.

But I do think there is some work here that needs to be done. Currently, clients only expire their sessions if all servers respond with a `NO_LEADER_ERROR`. But in reality, Copycat will never expire a session when no leader is present, and so clients sessions can persist through arbitrary leadership changes or lengthy periods of loss of quorum. What needs to be determined is when to expire the client's session in cases where it simply can't reach the cluster, the goal being to ensure that two Atomix clients with an open `LeaderElection` resource don't believe themselves to be the leader simultaneously (when one client can't send keep alive requests and so another client gets elected). It may be impossible to guarantee that because we have to rely on system clocks, but expiring clients' sessions when they fail to send keep alive requests is the next best option.

kuujo: Sorry you were talking about establishing a session, but I think the above also needs to be figured out. The point about what happens if a client can't initially establish a session we've debated in the past. When initializing the session, it does indeed currently only try each member once. I think failing the initial `open` call if a client can't immediately establish a session is intuitive. But there are definitely cases where it wouldn't be right to fail the `open` because a session couldn't be establish, e.g. when there's just no leader. So, one option is to add a strategy for opening a session. The `ConnectionStrategy`, which may be poorly named, dictates the servers to favor. In other words, for read-heavy workloads, connect to followers, and for write-heavy connect to the leader. I guess a strategy for opening a session would allow retries with exponential back off.

But we could also use the server response errors. If `NO_LEADER_ERROR` is returned by all servers, wait for a leader to be elected.

Maybe I'll just put all those in another strategy.

kuujo:
So, here are the potential consistency issues I see here. First, the issue of a client determining when it's session is expired requires some changes. That could be done in another PR. More relevant here are the configuration changes. Each server persists its last known *uncommitted* configuration to disk separate from the log, and each server uses its last known uncommitted configuration at startup. This follows the normal Raft model wherein configurations are applied when they're written to the log rather than when they're committed to aid the leader in determining when a configuration is actively in use, but only one configuration may be in the process of being committed at any given time. In Copycat's case, one of the potential concerns is that a server could be using a configuration that was never committed, and passive/reserve members don't participate in the commitment of configurations. But I believe this configuration and replication model - wherein leaders promote reserve to passive and passive to active and demote active to reserve - to be safe because the voting members are all that really matter for safety. The process of promoting and demoting active members which contribute to the quorum is still done according to the Raft specification. Only the processes of promoting reserve members and replicating to passive members from followers are different, but passive member replication relies on replicating only committed entries for consistency, and configurations on passive and reserve members are always synchronously updated by their related follower before sending `AppendRequest`s. Additionally, consistent hashing is done in the state machine to ensure partitioning and replication is consistent across followers.

This actually reminds me that there is a bug in the consistent hashing algorithm. Followers use consistent hashing to determine the reserve/passive members to which to send configuration updates and entries. But it doesn't currently account for the case where `quorumHint = 1`. If `quorumHint` is `1` the leader has to send configurations and entries to any reserve/passive member.

So, I think remaining bugs are:
* Clients' session expiration
* Promoting an active member is one is available *before* removing an active member that's leaving the cluster
* Replicating to reserve/passive members when `quorumHint` is `1`

Also, needs more tests for consistent hashing and new `*Appender` classes.

client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
madjam: Yep. That is what I had in mind as well. The existing `ConnectionStrategy` is more akin to a strategy for selecting a member to connect to. If we can support a strategy for initial session establishment, it would help clients configure what do if they go through all `connectMembers` and still could not establish a session. Retry after a delay is a simple one. We can perhaps toss in exponential back off as you suggested to ensure servers do not get overloaded while they are in the process of bootstrapping the Raft cluster.

madjam:
I have only managed to review may be the first 25% of files. I will be spending more time today to look at the rest. I haven't got to the point where Followers replicate log entries to Passive members. What is the rationale for using consistent hashing here? I get that we need to distribute the work load evenly among all followers. Is the "consistent" aspect required to ensure the same follower replicates log entries to the a given passive member even as the number of passive members change? 
Or may be I should wait until I look at that code :)

client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
madjam: Also, on the topic of connection establishment/selection/maintenance strategies, these are exposed at the copycat layer (and rightly so). But what if I am using AtomixClient?

madjam:
I'm jumping ahead here but wanted to confirm if the following is true: Does each active server have its own set of backups or do they form a common pool? The reason I ask is because there could be constraints regarding which passive member can take the place of which active server. For instance, you might not want all your active members in the same availability zone. So if a particular active member goes offline we want to preferably replace it another server that belongs to the same availability zone.

client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
kuujo: I think the idea for Atomix is to hide as much of the decision making as possible, but in some cases that may not be possible. For the `RecoveryStrategy` I submitted the PR to handle recovery of all resources in Atomix. I think the idea there is to just recover the resources and then expose an event interface to allow users to listen for events like session open/close/disconnected/etc (which haven't been added yet). For the connection related strategies, I suppose for simplicity we could just allow them to be set on `AtomixClient` for now.

kuujo:
Interesting point about AZs. So, currently what it does is just promote the most up-to date passive member if an active member goes down. It would be simple to categorize nodes though. Servers effectively identify themselves by sending their `clientAddress` to the leader in heartbeats. That address is logged and replicated in a configuration change. The same could be done in order to group nodes and only promote nodes within a group.

On a related note, while I agree that connection strategies and client communication patterns belong in Copycat, I'm not sure the concept of reserve nodes does. I'd love for only Atomix to support the `RESERVE` node concept, and that's really just a design problem. Now that the algorithm is done, I may explore whether reserve members can be moved to Atomix. The ideal scenario IMO would be for Copycat to only understand the concept of catching up `PASSIVE` servers and promoting them to `ACTIVE` and for Atomix to override that behavior to add `RESERVE` members and keep `PASSIVE` servers from being promoted until necessary, but keeping it in Copycat may (probably will) still turn out to be the concession that has to be made.

Then again, maybe I'm just being finicky about having it in Copycat.

kuujo:
I think it will likely be too difficult to cleanly separate the `RESERVE` state.

kuujo:
I think I'll be able to move some of this to Atomix. I'm actually going to close this and resubmit some of the final design improvements in separate smaller pull requests.

PR #64, opened 2015-11-27
Persist server term and vote

kuujo:
This PR implements a persistence layer for storing server term and last voted for according to the Raft spec. It adds a `MetaStore` class which is managed the `Storage` object. The `MetaStore` is a small file that stores server-related state.


madjam:
Looks good to me.
:+1:

PR #65, opened 2015-11-27
LeaderState#Replicator#commit() might return null if connection fails

bgloeckle:
`LeaderState#Replicator#commit()` might return null if connection fails in `#commit(MemberState, AppendRequest, boolean)`, or also if the `send(..)` call on the connection fails in `#commit(Connection,MemberState, AppendRequest, boolean)`. In that case the `commitFuture` might be completed in `commitTime(...)` and the `commitFuture` field might be reset to null (if `nextCommitFuture` is null) - in which case `#commit()` would return null.
Seems to be meaningful to return the (perhaps already completed) CompletableFuture.


kuujo:
Nice. Good catch!

PR #67, opened 2015-11-28
Separate client/server transports

kuujo:
This PR improves the client/server communication interfaces by supporting *optional* separate `Address`es for clients and servers. Each server can specify a client `Address` and `Transport` through which clients can communicate with the server. This will potentially allow for different types of transports and protocols in the future. The addition of client addresses necessitated some significant cleanup of the configuration system. In order to ensure client addresses are propagated, servers submit a `JoinRequest` to the leader which logs and replicates it as a normal configuration change. This means configuration changes now take place absent the actual quorum size changing as they're used to essentially identify nodes.

This PR should be reviewed by going through the commits sequentially. They should be easy to understand. Commits towards the end are mostly bug fixes.

### DIFF
--- a/client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
+++ b/client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
@@ -38,6 +38,7 @@ import io.atomix.copycat.client.response.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.ConnectException;
 import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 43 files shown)


kuujo:
BTW this also merges #64 so we should get that merged first.

kuujo:
The failing tests here are new. Have to take a look at those.

madjam:
I managed to look through all the commits and overall I am fine with the proposal. Since this is a big commit it was hard to cross check every piece of logic. But overall this seems reasonable to me.
From my side:  :+1: 
Sorry for the delay in reviewing this. Holidays and the large commit size are my excuses :)

kuujo:
Oh I know no worries. I didn't expect this to get reviewed very quickly. Since this is probably the last big PR and I don't want to make it any bigger, I think I'll merge it some time today after some more testing so we can move on to snapshots and smaller commits to clean up any remaining items. Once this and whatever comes of the snapshot PR are merged there will be a feature freeze until after a 1.0 release, so should be small commits - bug fixes and structural changes - from here on.

PR #69, opened 2015-11-30
Use volatile keyword for fields participating in double-checked locking

bgloeckle:
In order to force the compiler to not optimize access to a locally unchanged variable when entering a synchronized block, the variable needs to be defined as "volatile". See for example https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java.


kuujo:
Nice, good catch.

kuujo:
That test periodically fails on Travis for unknown reasons unrelated to this PR.

Issue #75, opened 2015-12-04
CompletableFuture might complete in unexpected thread - Copycat gets unreliable


bgloeckle:
Hi!

Copycat uses the class `CompletableFuture` heavily and expects the handlers along those "pipelines" to be called on the same thread as the `complete` function was called on: For example the server classes (e.g. ServerState, *State, ServerContext, ...) expect to be only called on the single thread that is created in a `SingleThreadContext` in the constructor of `ServerContext`. With this assumption, Copycat has much simpler code throughout these classes, because it does not need to take any care of multithreading. Jordan and I discussed about that already in a thread in the Google Group: https://groups.google.com/d/msg/copycat/p9j8I0SRw3M/xR7fwplvCwAJ.

Exactly the same example that I talk about in the mail thread came up for my implementation in diqube now again (diqube internally uses copycat to have a reliable way to distribute some internal data across the cluster).

In the logs of some of diqubes tests, the following lines started to show up sometimes, and in those cases the copycat server did not start up correctly:

```
23:42:20.629 [main] INFO  o.d.consensus.DiqubeCopycatServer [DiqubeCopycatServer.java:143] - Starting up consensus node with local data dir at '/home/diqube/.jenkins/workspace/diqube/diqube-server/data/consensus'.
23:42:20.632 [copycat-server-Address[/127.0.0.1:5101]] DEBUG i.a.c.server.storage.SegmentManager [SegmentManager.java:318] - Created segment: Segment[id=1, version=1, index=0, length=0]
23:42:20.634 [copycat-server-Address[/127.0.0.1:5101]] DEBUG i.a.c.s.s.ServerStateMachineExecutor [ServerStateMachineExecutor.java:199] - Registered value operation callback class org.diqube.cluster.ClusterLayoutStateMachine$GetAllTablesServed
23:42:20.634 [copycat-server-Address[/127.0.0.1:5101]] DEBUG i.a.c.s.s.ServerStateMachineExecutor [ServerStateMachineExecutor.java:199] - Registered value operation callback class org.diqube.cluster.ClusterLayoutStateMachine$GetAllNodes
23:42:20.634 [copycat-server-Address[/127.0.0.1:5101]] DEBUG i.a.c.s.s.ServerStateMachineExecutor [ServerStateMachineExecutor.java:190] - Registered void operation callback class org.diqube.cluster.ClusterLayoutStateMachine$RemoveNode
[...]

PR #76, opened 2015-12-05
ClientSession maintains wrong connectFuture if Client creates a connection ("#connect") too quickly.

bgloeckle:
See comment in source.

### DIFF
--- a/client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
+++ b/client/src/main/java/io/atomix/copycat/client/session/ClientSession.java
@@ -415,8 +415,19 @@ public class ClientSession implements Session, Managed<Session> {
     if (connectFuture == null) {
       // If there's no existing connect future, create a new one.
       LOGGER.debug("Connecting: {}", member);
+
+      boolean[] completedAlready = new boolean[] { false };
+
       connectFuture = client.connect(member).thenCompose(this::setupConnection).whenComplete((connection, error) -> {
+        // If the "connect" call is finished quickly, we might execute this piece of code before assigning the result of
+        // "whenComplete" to the connectFuture (line above).
+        // In that case we would first execute "connectFuture=null" and then after that 
+        // "connectFuture = client.connect(..." -> we'd have a connectFuture although we have connected already, all
+        // future calls to this #request(..) method will end up not opening new connections.
+        // We therefore maintain the completedAlready[] boolean which will be true in case the connection finished in
+        // one swoop and we set connectFuture = null in that case again (see below).
         connectFuture = null;
+        completedAlready[0] = true;
         if (!checkOpen || isOpen()) {
           if (error == null) {
             request(request, connection, future, checkOpen, recordFailures);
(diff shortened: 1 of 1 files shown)


kuujo:
That test failure is unrelated. Still have yet to reproduce whatever is going on in that test.

kuujo:
So, it looks like this can happen when `setupConnection` returns `CompletableFuture.completedFuture` right? That will occur when a `RegisterRequest` is being sent (no session is open yet), causing the future to be synchronously completed and then `null` to be assigned before the `connectFuture` is assigned.

But I think maybe this should be gone about another way. The semantics of threading in Copycat guarantee that asynchronous calls will not complete before the `connectFuture` is assigned. So, `setupConnection` should really just complete the future asynchronously to guarantee that the result of `setupConnection` is *always* asynchronous. This makes for a more consistent, easier to debug system. Copycat has some utilities for this:
```java
return Futures.completedFutureAsync(null, context.executor());
```

Ensuring that the behavior of each method is always consistent with regard to being synchronous or asynchronous ensures that these types of bugs don't arise in any other/future calls to the same method.

bgloeckle:
Yes, that's right, my implementation of Client did return an already completed CompletableFuture. The workaround I now implemented basically is the same as your suggested solution by using `Futures.completedFutureAsync`: https://github.com/diqube/diqube/blob/master/diqube-consensus/src/main/java/org/diqube/consensus/internal/DiqubeCatalystClient.java#L60.

Basically, this PR is about the same issue for which I created #75 - copycat does depend on *when* (and in *what* thread) the CompletableFutures are completed. This means it depends on specific timings. `Client#connect` now is already called in the correct ThreadContext (and therefore on the correct thread), but then there is the next invariant that the timing needs to be correct - something that cannot be seen easily.
IMHO depending on correct timings is typically pretty problematic, as timings can easily be implemented wrong (like e.g. when implementing some custom code to work with a component, but also when extending the component itself), but bugs are hard to find (because it usually works "most of the time" and debugging is hard). I think, a component should always ensure that it can handle all timings used by source code "outside" of the component. We though could debate if the `Client` implementation is actually "outside" of copycat/catalyst.

The solution provided in this PR is fairly simple and it would spare guys who want to implement their own Connection/Client/... quite some headaches (and I guess that might be a lot of guys, as at least most people implementing enterprise software will not want to use unsecured Netty connections, but will rather wrap the calls by secured Servlet-calls etc.). It took me I guess several hours to identify the problem. 
If though you think it is not meaningful to merge this PR, at least the JavaDoc of `Client#connect` should be adjusted to point out that the returned future has to be completed *asynchronously* on the current thread context and/or a general documentation for guys who want to use copycat for their implementations should be created, that points out that e.g. "all returned CompletableFutures always have to be completed on the correct ThreadContext (on the current one, if none is passed to the function) and always have to be completed asynchronously" (only if that statement is true, of course).

kuujo:
The client has been completely rewritten to be more testable and better manage connections since this was opened. Reopen if this issue is present in the new client.

PR #77, opened 2015-12-06
Support snapshotting state machines

kuujo:
This PR adds support for snapshotting state machines. Snapshots are supported via an interface that can be implemented by state machines:

```java
public class MyStateMachine extends StateMachine implements Snapshottable {
  private long counter;

  public void snapshot(SnapshotWriter writer) {
    writer.writeLong(counter);
  }

  public void install(SnapshotReader reader) {
    counter = reader.readLong();
  }

  public void increment(Commit<Increment> commit) {
    counter++;
    commit.close();
  }

  public void decrement(Commit<Decrement> commit) {
    counter--;
    commit.close();
  }

  public long get(Commit<Get> commit) {
    try {
      return counter;
    } finally {
      commit.close();
    }
  }

}
```

Snapshots are taken whenever the log becomes "compactable." Compactability is defined by the ability to rewrite a segment according to the log's configuration. Taking a snapshot results in a file being stored on disk. Once the file is stored on disk, all events sent to all sessions must be acknowledged up to the index at which the snapshot was taken before the snapshot can be marked *completed* and made available for recovery and replication. Once a snapshot is completed, a flag is set in the snapshot descriptor indicating that it's safe to use, and a `snapshotIndex` is set in the log `Compactor` indicating that `SNAPSHOT` commands can be removed up to the snapshot index during compaction.

Snapshots are replicated via the `InstallRequest`/`InstallResponse`. When a follower's `nextIndex` decreases to the leader's last snapshot index, the snapshot will be sent to the follower in chunks. Followers treat `InstallRequest` like normal `AppendRequest` and reset their election timers to ensure elections don't timeout while sending large snapshots.

Note that this implementation supports a mixture of snapshotting and log cleaning. The typical state machine will just use snapshots, but internally snapshots simply result in normal `clean`ing of entries from the log, and state machines can even `clean` some commits and snapshot others.


kuujo:
This PR actually got mixed with the other big PR. Going to reopen it.

PR #86, opened 2015-12-12
Recalculate completeIndex only after entries have been applied

kuujo:
This PR fixes a race condition in the internal `ServerStateMachine` when calculating the `completeIndex`. The problem is, the `completeIndex` was being calculated in the server thread prior to the command being actually applied to the state machine. This meant that state machine events were published *after* the `completeIndex` was calculated. This could result in `completeIndex` being improperly decreased if later calculated after events were published by the state machine.


jhalterman:
Fixes https://github.com/atomix/atomix/issues/80.

PR #89, opened 2015-12-13
Add simpler interface for specifying command types

kuujo:
This PR modifies the `Command` interface to provide an alternative, more intuitive method for specifying state machine command types. By default all commands are of the type `Command.Type.DEFAULT`. The default command type is a special type that changes behavior according to the state machine type. If the state machine implements `Snapshottable`, the command will be a `SNAPSHOT` type command. If the state machine does not implement `Snapshottable`, it will be treated as a `SEQUENTIAL` command to ensure tombstones are handled properly during log compaction.

A special command and compaction type - `UNKNOWN` - supports state machines which may use a mixture of storing command state in snapshots and cleaning commands from the log. The `UNKNOWN` compaction mode allows state machines to treat commands in whatever way they want, but that comes at a price. Commands can never be removed from the log until a snapshot is taken and the command is stored on all servers, and `UNKNOWN` commands must be compacted sequentially only during major compaction.


kuujo:
I actually ended up simplifying the `CompactionMode` interface and removing the `TypedCommand` interface for now. The default behavior of commands will still be safe as it pertains to the state machine and log compaction. But I'd like for some time and usage to inform solutions for simplifying the way we expose log compaction to the user. For now, the `CompactionMode` enum will remain.

PR #99, opened 2015-12-22
Handle duplicate vote requests from the same server

kuujo:
This PR properly handles duplicate vote requests from a single server. Previously, if a follower received a vote request from the same server for the same term twice, a consistency check in the `setLastVotedFor` method would result in an `IllegalStateException`. This fixes that issue by not calling that method if a vote request is from the candidate that was already voted for.


kuujo:
Tests fail due to client issues which are being resolved in a separate PR.

Issue #97, opened 2015-12-18
Intermitttent unit test failure in ClusterTest


madjam:
I observe intermittent unit test failures ( > 50%) while running local builds of Copycat. `ClusterTest::testMany` is the one that fails.

I'm at commit 310383bd9a15b670cfd42d7fe77aaada15bea27a

```Tests run: 84, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 476.549 sec <<< FAILURE!
testMany(io.atomix.copycat.test.ClusterTest)  Time elapsed: 40.868 sec  <<< FAILURE!
java.util.concurrent.TimeoutException: Test timed out while waiting for an expected result
	at net.jodah.concurrentunit.Waiter.await(Waiter.java:168)
	at net.jodah.concurrentunit.Waiter.await(Waiter.java:142)
	at net.jodah.concurrentunit.ConcurrentTestCase.await(ConcurrentTestCase.java:105)
	at io.atomix.copycat.test.ClusterTest.testMany(ClusterTest.java:135)```

I haven't done a lot of debugging yet. Thought I will log an issue in case some one runs into the same thing or have already figured out whats causing it.

kuujo:
Ditto.

@madjam here's my experience: I've actually been looking at this for a few days and am struggling to find any issue in Copycat. The interesting thing is, all tests seem to pass 100% of the time when run individually as far as I can tell, but when the cluster tests are run together one or more frequently fail. What I see in most failures is random leader elections happening. What seems to happen is the leader is blocked for several seconds and the followers election timeouts expire. I spent several hours trying to figure out what was blocking, but it seems to be nothing at all, or at least I couldn't find the culprit. It seemed to be blocking at random points that don't do any I/O.

I did do some profiling and noticed that some clients aren't being closed in test cleanup. Indeed, until recently the Atomix tests were failing because threads weren't being cleaned up, and that could also be the case here. But my brief efforts to make sure clients were closed were unsuccessful. YourKit still shows some client threads continuing.

kuujo:
[...]

PR #100, opened 2015-12-30
Client rewrite

kuujo:
This PR is a complete rewrite of the Copycat client to address test issues mentioned in #97. Previously, the client took a naive approach to submitting and resubmitting commands to the cluster and recovering lost sessions. The client would resubmit commands across sessions with sequence numbers for the original session, thus leading effectively to a "live lock" wherein the cluster could not progress because of improper command sequencing. Additionally, the original client attempted to use time to expire its own session.

The improved client present in this PR is a complete rewrite of the `CopycatClient`. The client's internals were separated into multiple separate, testable classes and tests with mocking were written for each of the internal helper classes. The new client tracks and exposes client state information in a manner that allows users to react to the client's state. It adds a `CopycatClient.State` enum which indicates whether the client is `CONNECTED`, `SUSPENDED`, or `CLOSED`. The `SUSPENDED` state is the most notable state in that it indicates that the client is unable to contact the cluster, though its session may not necessarily be lost. While the client is in the `SUSPENDED` state, it will continue to attempt to contact the cluster and keep its session alive. Once the client reestablishes communication with the cluster, it will determine whether its session was expired and, if so, can recover its session via a `RecoveryStrategy`.

Additionally, this client handles resubmitting commands in the client's program order across sessions quite gracefully. In the event that the client's session expires and the client must reopen a new session, the first action taken once a new session is opened will be resubmission of the pending commands from the prior session.

This PR improved the reliability of Copycat's cluster tests significantly. The goal of this PR, of course, is to completely and reliably eliminate test failures. Thus far, it seems all remaining test failures are actually minor issues with how the test `LocalTransport` handles some failures.


kuujo:
There seems to be an issue with one remaining test, the leader kill test. But seeing as how this is a huge improvement over the current `master` branch, I'm merging this and will address remaining issues separately.

PR #108, opened 2016-01-02
Fix bad condition in resending session events

kuujo:
The logic for resending events upon the failure of a client to receive an event has the potential to prevent session events from progressing altogether. The current code has the following condition:
```java
if (index > completeIndex) {
  // Resend events...
}
```
This results in the session only resending events when the client acknowledges a *new* event index. If the client fails to receive an event *after* it has already acknowledged the previous event index and the client requests events after that index, the events will never be resent since the `index > completeIndex` condition is always `false`. This PR modifies that logic to *always* clear acknowledged events and resend pending events. This may not totally be efficient as some pending events may already be in transit and so the session may send duplicate events, but it ensures that the event system progresses which is what's most important for this PR.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/ServerSession.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/ServerSession.java
@@ -566,11 +566,9 @@ class ServerSession implements Session {
    * @return The server session.
    */
   ServerSession resendEvents(long index) {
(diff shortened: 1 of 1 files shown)


kuujo:
This is the bug I've been looking for forever! :-) Pass and merge.

PR #157, opened 2016-02-18
Close expired session when client is successfully recovered.

### LINKED ISSUE #156: Expired session still left open after its recovery
When a expired session is recovered by opening a new session, we still leave the old session open. As result the old `ClientSessionManager` keeps sending periodic `KeepAliveRequest`s (which fail with `UNKNOWN_SESSION_ERROR` as expected) I noticed that this also prevents the client (whose session was recovered) from successfully submitting new operations possibly because it is still using a stale `ClientSessionSubmitter` (I'm not sure of this part).

I should have a pull request with the bug fix. It worked in my limited testing. But I'm not fully versed with the ClientSession recovery logic yet. So there could be some missing pieces.

madjam:
Fixes #156


kuujo:
Awseome! :+1:

kuujo:
I'm gonna merge this because I think it was causing some issues in my own testing :-)

Issue #158, opened 2016-02-19
Lagging follower is forced to truncate its log repeatedly


madjam:
I noticed this during a test run. This is a 3 node set up. One of the follower keeps resetting itself (truncate log) as it always discovers that the `globalIndex` is greater than what it is aware of. A brief look at the logic in `PassiveState::checkGlobalIndex` tells me that this should not happen on start up.

Here are some log lines. (This pattern keeps repeating)

```
2016-02-18 23:18:49,572 | DEBUG | FollowerState | 10.254.1.207/10.254.1.207:9876 - Received AppendRequest[term=1, leader=184429489, logIndex=0, logTerm=0, entries=[0], commitIndex=3288, globalIndex=3288]
2016-02-18 23:18:49,572 | DEBUG | SegmentManager | Closing segment: 1
2016-02-18 23:18:49,573 | DEBUG | SegmentManager | Created segment: Segment[id=1, version=1, index=0, length=0]
2016-02-18 23:18:49,573 | DEBUG | FollowerState | 10.254.1.207/10.254.1.207:9876 - Sent AppendResponse[status=OK, term=1, succeeded=true, logIndex=0]
2016-02-18 23:18:49,573 | DEBUG | FollowerState | 10.254.1.207/10.254.1.207:9876 - Received AppendRequest[term=1, leader=184429489, logIndex=0, logTerm=0, entries=[0], commitIndex=3292, globalIndex=3292]
2016-02-18 23:18:49,573 | DEBUG | SegmentManager | Closing segment: 1
2016-02-18 23:18:49,573 | DEBUG | SegmentManager | Created segment: Segment[id=1, version=1, index=0, length=0]
2016-02-18 23:18:49,573 | DEBUG | FollowerState | 10.254.1.207/10.254.1.207:9876 - Sent AppendResponse[status=OK, term=1, succeeded=true, logIndex=0]
[...]

PR #160, opened 2016-02-19
Reset follower state if globalIndex is greater than last log index

kuujo:
This PR fixes a bug associated with #158. When a follower checks if the `globalIndex` received from a leader is within the bounds of the follower's log, it currently does so using `contains`. However, `contains` excludes released and/or compacted entries in the log. Instead, this modification compares `globalIndex` against the log's `lastIndex` to simply check that the `globalIndex` is within the bounds of the log.


madjam:
Looks good.
:+1:

kuujo:
Not sure if this resolves the issue in #158 but it should have helped. We should keep evaluating that issue.

PR #169, opened 2016-02-24
Include cluster name in thread name for better diagnostics

madjam:
When running multiple embedded copycat clusters on the same node it very useful to be able identify where individual log messages are coming from.


kuujo:
I concur!

:+1:

PR #171, opened 2016-02-25
squid:S1213 - The members of an interface declaration or class should appear in a pre-defined order

georgekankava:
This pull request is focused on resolving occurrence of Sonar rule
squid:S1213 - The members of an interface declaration or class should appear in a pre-defined order.
You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S1213
Please let me know if you have any questions.
George Kankava

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/storage/snapshot/SnapshotDescriptor.java
+++ b/server/src/main/java/io/atomix/copycat/server/storage/snapshot/SnapshotDescriptor.java
@@ -32,6 +32,22 @@ import io.atomix.catalyst.util.Assert;
 public final class SnapshotDescriptor implements AutoCloseable {
   public static final int BYTES = 64;
 
+  private Buffer buffer;
+  private final long index;
+  private final long timestamp;
+  private boolean locked;
+
+  /**
+   * @throws NullPointerException if {@code buffer} is null
+   */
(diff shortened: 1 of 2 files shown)


kuujo:
The methods in Catalyst, Copycat, and Atomix do appear in a predefined order. The convention for method ordering is:
* static variables
* static methods
* constructors
* instance variables
* instance methods, generally ordered by dependencies (related/dependent methods neighboring each other)

This is the convention by which classes are ordered throughout all of the projects, and IMO consistency is more important than following some specific convention developed outside the scope of this project. Catalyst/Copycat/Atomix do indeed follow this convention *precisely*. To break this convention for a single class would be to break that convention. Either *all* of the classes need to be updated to follow this convention or none of the classes. But I will likely oppose a PR modifying this convention.

The reasoning behind placing static methods above constructors is because Catalyst/Copycat/Atomix use builders for which a static factory method is typically added to the class the builder constructs. For instance, `CopycatServer` has a `CopycatServer.Builder` which is constructed via the `CopycatServer.builder(...)` static methods. Placing the `builder` methods at the top of the class places the documentation of how to construct an instance of the class at the top of the class file. Moving the static `builder` factories below a series of often private constructors would move that documentation further down the class file below some code that's irrelevant to the user.

PR #173, opened 2016-02-26
Increase Copycat log entries length to be 32 bit

bgloeckle:
At the latest with https://github.com/atomix/catalyst/pull/30 copycat should support log entries that have a length that takes up full 32-bit, instead of cutting off at 16 bit.
Even without the above linked PR for catalyst, copycat still should support large entries, since users could install their own TypeSerializers which serialize to byte arrays that have a length larger than what can be displayed using 16 bit.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/storage/Segment.java
+++ b/server/src/main/java/io/atomix/copycat/server/storage/Segment.java
@@ -293,7 +293,7 @@ public class Segment implements AutoCloseable {
     boolean skipTerm = term == lastTerm;
 
     // Calculate the length of the entry header bytes.
-    int headerLength = Bytes.SHORT + Bytes.LONG + Bytes.BOOLEAN + (skipTerm ? 0 : Bytes.LONG);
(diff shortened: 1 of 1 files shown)


kuujo:
I'm good with this, just need to fix some of the log tests. Let me look in to them...

kuujo:
@bgloeckle https://github.com/bgloeckle/copycat/blob/fix-entry-length-in-log/server/src/main/java/io/atomix/copycat/server/storage/Segment.java#L82 and https://github.com/bgloeckle/copycat/blob/fix-entry-length-in-log/server/src/main/java/io/atomix/copycat/server/storage/Segment.java#L90 need to be changed to `readInt` as well. When a segment is first created, if the segment is persisted on disk then it rebuilds the index, reading the length of each entry sequentially.

I think this PR should be good after that.

kuujo:
@bgloeckle nevermind I saved you some time and updated and merged it locally

bgloeckle:
Ah, thanks for taking care of that! I missed it. If I can be of any help with this, just let me know :)

kuujo:
:+1: thanks!

PR #174, opened 2016-02-27
Ensure entries released during minor compaction are transferred to compacted segments

kuujo:
This PR resolves the same issue that was previously present in major compaction wherein some entries released from a segment *after* the start of compaction of that segment are not properly transferred to the compacted segment. This simply updates the minor compaction process to copy all cleaned entries after a segment has been compacted. This process should be extremely fast and efficient since only in-memory indexes are read sequentially which is an O(1) operation.


kuujo:
Additionally, some of the terminology Copycat uses to explain log compaction has changed over time. c20843b updates some of the class and method names and documentation to ensure the terminology used is consistent with documentation.

PR #181, opened 2016-03-08
Ensure a leader can replicate log entry larger than MAX_BATCH_SIZE

madjam:
This fixes an issue where when a large entry (size >= MAX_BATCH_SIZE) is encountered, the Leader append entries logic keeps repeatedly sending a empty batch request (in a tight loop)


kuujo:
Awesome! :+1:

PR #182, opened 2016-03-09
Ensure single-node cluster does not attempt to join

kuujo:
Single nodes in a single node cluster will attempt to join a cluster that doesn't exist. This adds a condition to check the active members size and complete the join immediately for single-member clusters.


jhalterman:
:+1:

PR #186, opened 2016-03-11
Automatically fail non-timeout or response exceptions

kuujo:
This PR modifies the logic for handling exceptions in the client to immediately fail all exceptions other than `Response` errors or `TimeoutException`. Additionally, it ensures that such exceptions bubble up to the user code so users can properly react to serialization failures in particular.


kuujo:
Added `ConnectException` to the list of exceptions that are supported for retry.

Issue #185, opened 2016-03-11
IndexOutOfBoundsException on a Log::get


madjam:
Noticed this with latest copycat. Logging an issue so that I don't miss it.
```
2016-03-06 21:36:18,479 | ERROR | 9876-partition-3 | SingleThreadContext              | 68 - org.onosproject.onlab-thirdparty - 1.5.0.SNAPSHOT | An uncaught exception occurred
java.lang.IndexOutOfBoundsException
	at io.atomix.catalyst.buffer.util.BitArray.get(BitArray.java:97)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.storage.util.OffsetPredicate.test(OffsetPredicate.java:53)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.storage.Segment.isLive(Segment.java:442)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.storage.Log.get(Log.java:371)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.AbstractAppender.getPrevEntry(AbstractAppender.java:220)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.AbstractAppender.buildAppendEntriesRequest(AbstractAppender.java:164)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.AbstractAppender.buildAppendRequest(AbstractAppender.java:134)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.LeaderAppender.appendEntries(LeaderAppender.java:206)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.LeaderAppender.appendEntries(LeaderAppender.java:110)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.LeaderState.appendMembers(LeaderState.java:143)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
[...]

PR #191, opened 2016-03-14
Fix IndexOutOfBoundsException when checking OffsetPredicate for offse…

kuujo:
…t that has not been released.

#185

@madjam thoughts?


madjam:
Looks good to me :+1

PR #193, opened 2016-03-17
Check for retryable cause when error is a  CompletionException

madjam:
`CompletableFuture::join` can throw a `CompletionException` and when that happens subsequent completion stages complete with a `CompletionException` as well. For retry logic to correctly work we need to check the cause when error is a `CompletionException`

This PR also reverts a change that was accidentally pushed to master branch.


kuujo:
:+1:

PR #196, opened 2016-03-21
Set command and query errors as retryable

madjam:
We currently fail on `QUERY_ERROR` and `COMMAND_ERROR`. These errors are encountered when the leader fails to get a quorum. However, these errors are potentially recoverable. This PR will (re-)enable retries for these errors.


kuujo:
Ahh you're right. I think those used to also be returned when the state machine threw an exception, so the purpose of those checks was to prevent a command that causes an exception in the state machine from being repeatedly resubmitted. But I double checked and that appears to no longer be the case. 

That's my long way of saying :+1:

PR #198, opened 2016-03-23
Session related bugfixes + logging improvements

madjam:
This PR addresses the following issues:
- Ensure server side session state is appropriately cleaned up on session expiration. This helps prevent invalid session state transitions.
- On the client side do not schedule a recurrent keep alive task up front. This helps prevent sending keep alives for sessions the server no longer recognizes.
- Misc logging improvements for better diagnostics

### DIFF
--- a/protocol/src/main/java/io/atomix/copycat/protocol/KeepAliveResponse.java
+++ b/protocol/src/main/java/io/atomix/copycat/protocol/KeepAliveResponse.java
@@ -123,7 +123,7 @@ public class KeepAliveResponse extends SessionResponse {
 
   @Override
   public String toString() {
-    return String.format("%s[status=%s, leader=%s, members=%s]", getClass().getSimpleName(), status, leader, members);
+    return String.format("%s[status=%s, error=%s, leader=%s, members=%s]", getClass().getSimpleName(), status, error, leader, members);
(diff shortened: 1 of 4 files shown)


protocol/src/main/java/io/atomix/copycat/protocol/KeepAliveResponse.java
```diff
@@ -123,7 +123,7 @@ public boolean equals(Object object) {
 
   @Override
   public String toString() {
-    return String.format("%s[status=%s, leader=%s, members=%s]", getClass().getSimpleName(), status, leader, members);
+    return String.format("%s[status=%s, error=%s, leader=%s, members=%s]", getClass().getSimpleName(), status, error, leader, members);
```
kuujo: Maybe we should add the `error` to all the other responses too?

madjam: I agree. Having this field logged helped a lot. Shall we do that in another PR?

kuujo: Sounds good

kuujo:
:+1:

PR #199, opened 2016-03-24
Fixes bug that was preventing client recovery more than once

madjam:
This PR fixes a problem that was limiting a CopycatClient's ability to recover to only one time.


kuujo:
Nic :+1:

PR #200, opened 2016-03-26
Ensure EventHolder previousIndex is >= completeIndex

madjam:
This PR fixes a issue that was identified with event publishing around leadership transfers. If the new leader has previously not published any events or if it has not even buffered any of the previously published events, the first event it publishes (after assuming leadership) will potentially have a `previousIndex` that is smaller than the last received event index (represented by `completeIndex`). When this condition occurs the event publishing to that client locks up.


kuujo:
Yay! :+1:

PR #203, opened 2016-03-28
Performance improvements

kuujo:
This PR adds various performance improvements to logging and replication in Copycat. The performance improvements are primarily with memory mapped files. `flush`ing files to disk was moved to ensure it only occurs when the commit index is increased rather than on each write. This allows memory mapped files to provide very fast random access to writes while still maintaining consistency flushing to disk prior to completing writes. I also experimented with pipelining in this branch, however no simplistic implementation of pipelining performed better than simple batching, and indeed in most cases it performed worse due to significantly smaller batches.

In my tests, I've seen a cluster of three nodes with `MAPPED` storage receive up to 60k writes/sec from a single client.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
@@ -112,7 +112,7 @@ abstract class AbstractAppender implements AutoCloseable {
     final long index = prevEntry != null ? prevEntry.getIndex() + 1 : context.getLog().firstIndex();
 
     // Build a list of entries to send to the member.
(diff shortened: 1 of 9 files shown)


kuujo:
One more note: Memory mapped segments were also changed so the first time a memory mapped segment is compacted, the segment is converted to a disk based segment. Older segments that have been compacted should not consume extra resources as they're never written and are only read when catching up a new or far behind server.

madjam:
:+1: 

For pipelining I wonder if the communication latency between nodes is too small and therefore additional serialization overhead due to small batches starts to dominate.

kuujo:
@madjam I had that thought too. I'm going to open a separate PR for pipelining and run it on EC2 to see what the result is.

PR #205, opened 2016-03-29
Implement pipelining

kuujo:
This PR is an initial implementation of pipelining on Copycat. I took two approaches to attempting pipelining. The first approach was a naive one in which leaders simply pipelined `n` number of requests. The assumption is that under high load, batch sizes will reach an equilibrium. The second approach that was attempted was one alluded to in the Raft dissertation. This approach attempted to estimate the round trip time for an `AppendRequest` to each individual member and use that to optimize batch sizes.

Ultimately, after significant benchmarks using both approaches, I decided to use the latter, more intelligent approach as it proved to be significantly faster than the naive approach in tests. To estimate optimal batch sizes, the round trip time for each request is stored in the `MemberState`, and the last `8` round trip times are averaged to determine when a batch should be sent. Once an append is attempted after 1/2 the average round trip time, pending entries will be batched and sent to the follower. Because of the way the logic is written, this should have no negative impact on write latency during period of low load. In the event a write is made under low load, the `AppendRequest` will immediately be sent since `System.currentTimeMillis()` will normally already be past the required half round trip time.

@madjam can you review?

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
@@ -191,11 +191,16 @@ abstract class AbstractAppender implements AutoCloseable {
    * Sends a commit message.
    */
   protected void sendAppendRequest(Connection connection, MemberState member, AppendRequest request) {
+    long timestamp = System.currentTimeMillis();
     connection.<AppendRequest, AppendResponse>send(request).whenComplete((response, error) -> {
       context.checkThread();
 
       // Complete the append to the member.
-      member.completeAppend();
+      if (!request.entries().isEmpty()) {
+        member.completeAppend(System.currentTimeMillis() - timestamp);
+      } else {
+        member.completeAppend();
+      }
 
       if (open) {
         if (error == null) {
@@ -206,6 +211,11 @@ abstract class AbstractAppender implements AutoCloseable {
         }
       }
     });
+
+    updateNextIndex(member, request);
+    if (!request.entries().isEmpty() && hasMoreEntries(member)) {
+      appendEntries(member);
+    }
   }
 
   /**
@@ -322,9 +332,11 @@ abstract class AbstractAppender implements AutoCloseable {
   /**
    * Updates the next index when the match index is updated.
    */
-  protected void updateNextIndex(MemberState member) {
+  protected void updateNextIndex(MemberState member, AppendRequest request) {
     // If the match index was set, update the next index to be greater than the match index if necessary.
-    member.setNextIndex(Math.max(member.getMatchIndex() + 1, 1));
+    if (!request.entries().isEmpty()) {
+      member.setNextIndex(request.entries().get(request.entries().size()-1).getIndex()+1);
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
```diff
@@ -191,11 +191,16 @@ protected void sendAppendRequest(MemberState member, AppendRequest request) {
...
    */
   protected void sendAppendRequest(Connection connection, MemberState member, AppendRequest request) {
+    long timestamp = System.currentTimeMillis();
     connection.<AppendRequest, AppendResponse>send(request).whenComplete((response, error) -> {
       context.checkThread();
 
       // Complete the append to the member.
-      member.completeAppend();
+      if (!request.entries().isEmpty()) {
+        member.completeAppend(System.currentTimeMillis() - timestamp);
```
kuujo: Record the round trip time

server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
```diff
@@ -206,6 +211,11 @@ protected void sendAppendRequest(Connection connection, MemberState member, Appe
         }
       }
     });
+
+    updateNextIndex(member, request);
```
kuujo: Optimistically update `nextIndex` and send pending entries if available

server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
```diff
@@ -322,9 +332,11 @@ protected void updateMatchIndex(MemberState member, AppendResponse response) {
   /**
    * Updates the next index when the match index is updated.
    */
-  protected void updateNextIndex(MemberState member) {
+  protected void updateNextIndex(MemberState member, AppendRequest request) {
     // If the match index was set, update the next index to be greater than the match index if necessary.
-    member.setNextIndex(Math.max(member.getMatchIndex() + 1, 1));
+    if (!request.entries().isEmpty()) {
+      member.setNextIndex(request.entries().get(request.entries().size()-1).getIndex()+1);
```
kuujo: Update `nextIndex` based on the last appended entry's index + 1. We have to use the last entry rather than the entry count since incremental compaction may exclude entries from replication.

server/src/main/java/io/atomix/copycat/server/state/MemberState.java
```diff
@@ -214,7 +218,7 @@ MemberState setNextIndex(long nextIndex) {
    * @return Indicates whether an append request can be sent to the member.
    */
   boolean canAppend() {
-    return !appending;
+    return appending < MAX_APPENDS && System.currentTimeMillis() - (timeBuffer.average() / MAX_APPENDS) >= appendTime;
```
kuujo: Whether an `AppendRequest` can be sent to a follower depends on two factors:
* Less than `2` requests are already outstanding
* More than half the average round trip time for the last 8 non-empty requests has elapsed since the start of the last request

server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
```diff
@@ -191,11 +191,16 @@ protected void sendAppendRequest(MemberState member, AppendRequest request) {
    * Sends a commit message.
    */
   protected void sendAppendRequest(Connection connection, MemberState member, AppendRequest request) {
+    long timestamp = System.currentTimeMillis();
```
madjam: How about using System.nanoTime() for better resolution?

madjam:
:+1: 
Looks good overall

server/src/main/java/io/atomix/copycat/server/state/AbstractAppender.java
kuujo: :+1:

kuujo:
`System.nanoTime` change added

madjam:
:+1:

PR #206, opened 2016-03-30
Multi threaded appender

kuujo:
This PR adds performance improvements to appenders in Copycat. Each appender has a thread pool, and each member operates on a single thread in the pool at any given time. `ThreadContext`s are multiplexed across the thread pool such that available threads are used rather than assigning each specific context to a specific thread. This is done with `ThreadPoolContext`. Creating a thread-per-follower has proved to increase the performance of replication after recent performance fixes in the `NettyTransport`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <java.version>1.8</java.version>
     <slf4j.version>1.7.7</slf4j.version>
     <logback.version>1.1.2</logback.version>
-    <catalyst.version>1.0.5</catalyst.version>
+    <catalyst.version>1.0.6-SNAPSHOT</catalyst.version>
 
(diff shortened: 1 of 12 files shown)


kuujo:
@madjam thoughts?

I sort of threw this together pretty quickly, but it showed some promise in tests.

madjam:
@kuujo This looks good to me. :+1: 
What performance gains did you see in your experiments?

kuujo:
@madjam actually, I realized I missed something here. It turns out that all the performance gains were in atomix/catalyst#44 and I forgot to bump the version in my tests against `master`. I went back and tested against master with the Netty performance fix, and this PR does actually degrade performance a bit. I think this should be closed.

kuujo:
We just need to bump the Catalyst version for the performance :-)

PR #209, opened 2016-04-04
Separate bootstrap and join APIs

kuujo:
This PR refactors the cluster configuration API to separate `bootstrap` and `join` methods. This will promote clarity for users on what occurs when a server is started. The current API which transparently reconfigures the cluster has proven to be too confusing to Copycat and Atomix users.


kuujo:
@madjam our thought about this was the current API is too confusing for users since it transparently bootstraps or joins a cluster. By providing separate methods for each, the behavior of starting a server becomes a lot more predictable and obvious. Thoughts?

madjam:
I'm definitely in favor of making this change. Agree that it makes the intent of the API obvious to the user.

kuujo:
Going to finish up a PR for Atomix before merging this

PR #210, opened 2016-04-08
Full sequential consistency for all operation responses and events

kuujo:
This PR implements sequential consistency for operation responses and events and fixes several bugs with sequencing operations in FIFO order on servers, in particular when clients switch between followers and leaders for reads and writes respectively. To implement sequencing for responses/events, we remove support for retrying client queries. Commands are still retried internally for linearizable semantics, but because failed queries don't have the potential for side effects and because queries can switch between the leader and followers, we found removing retries for queries altogether to be the simplest approach at least for the initial implementation. This forces the user to retry queries externally, implying multiple operations. In the future, we can at least make queries *more likely* to succeed by implementing retries for a single connection communicating with a single follower, but switching followers poses significant challenges for retrying queries and maintaining FIFO order.

The goal of this PR is to entirely replace linearizable events with sequential consistency for all events received by a client (including responses). Once linearizable events were removed, the algorithms with which servers sequence commands and queries in FIFO order were cleaned up to ensure the correct behavior. We assume that requests arriving on a follower will always arrive in FIFO order given an ordered protocol, so to apply a query on a follower we simply wait for the application of the prior command. In contrast, commands sent to leaders may be proxied through followers and therefore have the potential to arrive out of order when switching servers, and they're therefore sequenced on the leader before being written to the Raft log. Additional issues were resolved in how queries applied on leaders are sequenced as well. Previously, there was some potential that a query arriving after a command could be applied before it if the earlier command is not committed in a single heartbeat. Therefore, leaders now appropriately await the application of the prior command sequence number to the state machine before evaluating a query.

Finally, with command and query requests properly sequenced on servers, client-side sequencing was implemented to ensure that clients see command and query responses in the order in which their counterpart requests were applied to state machines and to sequence events received from the cluster within those responses. The [ClientSequencer](https://github.com/atomix/copycat/blob/sequential-events/client/src/main/java/io/atomix/copycat/client/session/ClientSequencer.java) works by first placing responses in the order in which they were submitted to the cluster. For each response, if the prior event has not yet been received, the response is enqueued until the relevant criteria are met. See the `ClientSequencer` documentation for a more in-depth explanation of the algorithm it uses.


kuujo:
Working on the broken tests, but this should still be reviewed.

kuujo:
Tests look good now :+1:

kuujo:
@madjam `testSequenceOperations` in `ClusterTest` is really the Rey that demonstrates sequencing. A command is written, then a command that publishes an event, then a query. What the client sees from the cluster is:
* Command response
* Command response
* Event
* Query response

kuujo:
Test*

madjam:
Just trying to understand the semantics:

Lets say there are two clients `A` and `B`, `A` submits a command and before the command completes submits a query. So at A, the sequence of events is:
* Command completes
* Event arrives
* Query completes.

If `B` submits a query after `A`'s command completes, is it always guaranteed to see the value written by `A`? Will `B` receive the event before its query completes?

kuujo:
B is guaranteed to see A's command if the query from B is `LINEARIZABLE` or `BOUNDED_LINEARIZABLE` (don't love the name of that), and B will therefore also see the event before its query completes. However, if B's query is `SEQUENTIAL` then its query may read prior to A's command. But once B does submit a query that reads A's command, it's guaranteed never to read prior to A's command when switching servers.

kuujo:
I see a bug here stills when a `SEQUENTIAL` query is submitted to the leader, the leader immediately applies the query. This means a query submitted to the leader immediately after a command can be applied before the command. The leader should be using the `PassiveState`'s query methods for sequential queries.

kuujo:
Fixed and renamed `BOUNDED_LINEARIZABLE` to `LINEARIZABLE_LEASE` which is more explicit and descriptive of what it actually does (uses the leader's election timeout to assume no other leaders can exist).

kuujo:
It looks like the threading model on the client may not be correct too. From what I'm reading, it looks like responses will be completed on the I/O thread and events on the event thread, but they should both occur on the same thread. Will tackle that in the morning.

kuujo:
On that note, it might also be nice to do what ZooKeeper does and complete blocking response futures on the I/O thread so users can block inside the event thread when a response or event is received. This should be simple to do by extending `CompletableFuture` to set a flag when `get()` or `join()` is called and check the flag prior to completing the response.

kuujo:
502df7e fixes the thread issue and implements support for blocking in event threads.

PR #217, opened 2016-05-10
Always allow new appends if no appends are in progress

madjam:
This PR fixes a bug where a leader could stall processing a `LINEARIAZBLE` query. This typically happens when a bunch of such queries are submitted one after another in quick succession.
A `LINEARIZABLE` query needs to perform a quick heartbeat to ascertain quorum. But due to the bug the corresponding `AppendRequest` cannot be sent out because `MemberState::canAppend` returns `false`. This leaves `heartbeatFuture` as non-null and subsequent heartbeat’s get queued up. Only way to exit this impasse if is if a new entry is appended to the log (A `KeepAlive` will do) and replicated. But if the followers timeout on leader heartbeat before that happens, they will start a new poll and one of them will take over leadership.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/MemberState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/MemberState.java
@@ -218,7 +218,7 @@ final class MemberState {
    * @return Indicates whether an append request can be sent to the member.
    */
   boolean canAppend() {
-    return appending < MAX_APPENDS && System.nanoTime() - (timeBuffer.average() / MAX_APPENDS) >= appendTime;
+    return appending == 0 || (appending < MAX_APPENDS && System.nanoTime() - (timeBuffer.average() / MAX_APPENDS) >= appendTime);
   }
 
   /**


kuujo:
I think this needs to be thought through a bit. I think this change may result in some undesirable behavior (not that the current behavior is not undesirable).

IIRC the reason for not including `appending == 0` in that condition is to ensure the leader doesn't always send batches of size `1` if no requests are outstanding. What we see with the `appending == 0` condition is a leader will always send an `AppendRequest` immediately when a write is submitted and no other writes are outstanding. Ideally, the leader should wait half the average round-trip time before sending that request. But this is conflicting with reads.

I'm wondering if we can just implement logic for `LINEARIZABLE` reads that circumvents this logic but retain the logic for writes. IIRC in my tests when the `appending == 0` condition was included I ended up getting smaller write batch sizes even though it seems they should balance out after the first 1-entry `AppendRequest` under high load.

madjam:
So basically this is a trade-off between latency and efficiency (indirectly on throughput) and I guess there is no one correct answer for all use cases. Having said that in my test under load I did not see a significant impact on batch size.

kuujo:
Right and I wouldn't expect there to be. In theory, once the first request sends one entry the second should wait half the round trip time and batching should become efficient at that point, so this is really just an optimization for the first non-concurrent write which can in practice be quite frequent. I was surprised that it did have an impact on my own tests.

I'm not really opposed to better latency for the first write as long as no throughput is lost with vs without this condition. I'll look in to this some more and run more tests again and I suspect I won't see what I was seeing before and then I'll merge it :-)

kuujo:
I'm seeing the same performance with both approaches as expected. I'm going to merge this so we can focus on better things than initial write latency :-)

PR #222, opened 2016-05-19
Fixes the sequencing logic so that failed queries can be skipped over

madjam:
This PR fixes how responses are sequenced in the face of query failures. When a query fails we should mark its sequence number as abandoned and move on so that we don't block subsequent operations. This change ensure that.


kuujo:
:+1: awesome!

PR #223, opened 2016-05-19
Fix state machine error handling

davidmoravek:


kuujo:
The failed test is one that has failed intermittently on Travis for a long time that I've never been able to fully resolve. Will rerun the tests.

PR #224, opened 2016-05-20
Refactoring to make Copycat OSGi compliant

madjam:
This PR is a necessitated by https://github.com/atomix/catalyst/pull/49


jhalterman:
Putting the Listener stuff in .concurrent seems like a mismatch, at least in name. Does it need to be in the same package as the .concurrent stuff or maybe something like an .event package?

kuujo:
Rerunning tests since atomix/catalyst#49 was merged

PR #227, opened 2016-06-03
Fixes client sequencing bug

madjam:
This PR fixes a bug in the client sequencing logic. For this bug to manifest responses need to arrive out of order *and* each response should have a different event index *and* the later arriving response should have a higher event index

Here is an example that causes this:

(starting `eventIndex` is `500`)
- Client sends request with seq `100`
- Client sends request with seq `101`
- Response arrives for request `101` with event Index `500` (gets added to queue)
- Event `501` arrives (gets added to queue)
- Response arrives for request `100` with event Index `501` (this completes event `501` and request `100`)

Request `101` can no longer be completed because `eventIndex` has now advanced to `501`.

### DIFF
--- a/client/src/main/java/io/atomix/copycat/client/session/ClientSequencer.java
+++ b/client/src/main/java/io/atomix/copycat/client/session/ClientSequencer.java
@@ -177,7 +177,7 @@ final class ClientSequencer {
       }
     }
 
-    // If after completing pending events the eventIndex is equal to the response's eventIndex, complete the response.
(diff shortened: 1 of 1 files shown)


kuujo:
Awesome! :+1:

kuujo:
I expect to be cutting a new Atomix release this weekend so I'll push this too since its critical.

kuujo:
Restarting tests...

kuujo:
Looks good.

madjam:
Awesome! This took some time to track down. But I'm glad we found this in time.

kuujo:
Thanks!

PR #231, opened 2016-06-14
Avoids ConcurrentModificationException in ClientSessionSubmitter::close

madjam:
This PR fixes a bug in `ClientSessionSubmitter::close` method that can lead to a `ConcurrentModificationException`. When this happens it impacts client session recovery.


kuujo:
Just to make sure I'm getting this, we're replacing `LinkedHashMap` with `HashMap` and then sequencing the exception to ensure they occur in the correct order?

madjam:
LinkedHashMap change is just a simple optimization. By providing an insertion order iteration over values it lets us abort attempts in the order in which they are initiated. 

The key change is avoidance of concurrent modification of the map during iteration.

kuujo:
👍

kuujo:
I did see some test failures but they seem unrelated and I'll try to reproduce them separately.

madjam:
Great. Sounds good.

PR #234, opened 2016-06-16
Add expire() method to ClientSession for testing

kuujo:
This PR adds a simple `expire()` method to `ClientSession` for testing. When a session is forcefully expired, the keep-alive is cancelled and the session's state is set to `EXPIRED`. This is enough to cause `RecoveryStrategy` to be called.


madjam:
👍

PR #235, opened 2016-06-23
Ensure duplicate PublishRequests are handle appropriately

madjam:
This PR addresses stability issues that were being observed when several concurrent updates are made to a resource and where each such update generates a event notification. The issue stems from the fact that events can be delivered out of order and when that happens the out of sequence event gets rejected by the client (with `ERROR` status). This forces the server to immediately resend another large batch of events that are potentially in flight already. In my testing I have observed that this causes a flood or messages forcing new timeouts and more out of order delivery and thereby putting the system into a tail spin.


kuujo:
Make sense. Fix shouldn't cause any problems... `testSessionIgnoresOldEvent` needs to be updated for the response status change.

madjam:
Doh! I should have run the unit tests first :)

kuujo:
👍

PR #240, opened 2016-07-07
Fix recovery state transition

dmvk:


JozoVilcek:
@kuujo Any plans to get this released? Thanks.

PR #244, opened 2016-08-09
Use interval instead of delay to reschedule

### LINKED ISSUE #243: Scheduling recurrent tasks does not work as expected
ServerStateMachineExecutor reschedules recurrent tasks according to delay rather than interval.

``getExecutor().schedule(Duration.ofSeconds(1), Duration.ofSeconds(5), task);``

`task` gets called every 1 second rather than every 5.

JPWatson:
Fixes #243


JPWatson:
Looks like the build failure is due to a timeout rather than my change...

kuujo:
Yep, that issue was fixed in another PR.

PR #247, opened 2016-09-06
Fix 100% cpu utilisation on single CPU

JPWatson:
Number of compaction threads must be more than 0 otherwise scheduledThreadPool is initialized with 0 threads and you get 100% cpu spin.


madjam:
👍

madjam:
We may need a similar fix in Atomix (io.atomix.manager.options.ServerOptions)

PR #249, opened 2016-09-20
fix loosing exception cause

mnicky:
Wrong constructor was being called and the exception cause was effectively being lost.


kuujo:
Need to check other exceptions to ensure this isn't an issue elsewhere

PR #258, opened 2016-12-23
ServerSessionManager: only try to remove consistent information

### LINKED ISSUE #257: Session unregister might destroy other session information
Suppose a scenario where the following happens:

1) A client X tries to register a session and it times out. The session, however, was registered on server side. Let's say this was session 3.
2) Client X tries again to register a session and it works. It gets session 17.
3) Some time later, the abandoned session 3 expires and is going to be removed. Ultimately `ServerSessionManager:unregisterSession` is going to be called.

```
ServerSessionContext unregisterSession(long sessionId) {
    ServerSessionContext session = sessions.remove(sessionId);
    if (session != null) {
      clients.remove(session.client());
      addresses.remove(session.client());
      connections.remove(session.client());
    }
    return session;
  }
```

The information on those maps is already for session 17 and it is going to be destroyed.

thiagoss:
It is possible that a client creates 2 sessions if the first attempt
times out. In this case the ServerSessionManager will associate the 2nd
session to the client but the 1st session that is left unatended will
time out and eventually be unregistered.

When this is done it will cause the maps in this object to have the
client/session entry removed but those are already pointing to the new
information that is in use. Their removal will cause issues in cluster
operation.

Fixes #257

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/ServerSessionManager.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/ServerSessionManager.java
@@ -126,9 +126,9 @@ class ServerSessionManager implements Sessions {
   ServerSessionContext unregisterSession(long sessionId) {
     ServerSessionContext session = sessions.remove(sessionId);
     if (session != null) {
-      clients.remove(session.client());
-      addresses.remove(session.client());
-      connections.remove(session.client());
+      clients.remove(session.client(), session);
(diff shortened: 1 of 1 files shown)


kuujo:
This is great! I was actually just looking at this yesterday and realized the client ID should probably also be used to unregister clients' sessions when a new session for the client is registered as well to ensure old sessions don't hang around for too long, but that should be done in another issue.

PR #260, opened 2016-12-23
Reject messages from clients that are connected to the wrong server

### LINKED ISSUE #259: Race at client connection can lead to invalid client vs cluster state
A client can try to reconnect to the cluster after a disconnection/timeout and it might lead to an inconsistent state:

1) Client tries to connect to Server A and times out
2) Client tries to connect to Server B and succeeds
3) It is published that Client is connected to Server B
4) The first attempt at 1 finishes its process and it is published that Client is connected to Server A

In the end, the Client will be connected  to Server B but the cluster thinks it is connected to Server A. So no information about the events or replies to queries will ever be sent to the Client because Server B that is responsible for it has no connection. I have logs in case it is needed to confirm the issue.

thiagoss:
This will make them eventually timeout and try a new server instead of
trying to use a server that they aren't supposed to be using.

This can happen on races of clients connecting to multiple servers
due to a sequence of timeouts (frequent on high-load scenarios) and
the ConnectRequest/AcceptRequests ending up on the Leader at the
wrong order.

Example:

1) Client A tries to connect to Server X and times out
2) Client A tries to connect to Server Y and succeeds
3) Leader receives AcceptRequest from Server Y and processes/publishes it
4) Leader receives AcceptRequest from Server X and processes/publishes it

After this, Client A is connected to Server Y but the cluster believes
it should be connected to server X. This causes some inconsistency
checks that eventually leads to the Client A becoming useless and having
all its operations timeout. Its internal session state won't receive
publishes because Server Y won't send them (its connection was closed
once it received the message that Client A is now connected to Server
X).

The solution on this patch is to reject messages from clients that are
not connected to a particular server, leading the client to eventually
timeout and start a new connection. A new field was added to Session
requests to allow identifying when a request was forwarded (must be
handled) or when it came directly from a client (needs check if the
client is connected to the receiver).

Fixes #259

### DIFF
--- a/protocol/src/main/java/io/atomix/copycat/protocol/SessionRequest.java
+++ b/protocol/src/main/java/io/atomix/copycat/protocol/SessionRequest.java
@@ -30,6 +30,7 @@ import io.atomix.catalyst.util.Assert;
  */
 public abstract class SessionRequest extends AbstractRequest {
   protected long session;
+  protected boolean forwarded;
 
   /**
    * Returns the session ID.
@@ -42,12 +43,24 @@ public abstract class SessionRequest extends AbstractRequest {
 
   @Override
   public void readObject(BufferInput<?> buffer, Serializer serializer) {
+    super.readObject(buffer, serializer);
     session = buffer.readLong();
+    forwarded = buffer.readBoolean();
   }
 
   @Override
   public void writeObject(BufferOutput<?> buffer, Serializer serializer) {
+    super.writeObject(buffer, serializer);
     buffer.writeLong(session);
+    buffer.writeBoolean(forwarded);
+  }
+
+  public void setForwarded() {
+    forwarded = true;
+  }
+
+  public boolean wasForwarded() {
+    return forwarded;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


thiagoss:
I'm still not entirely sure about this patch but it gives an idea of what can be done to fix this issue. If you have a better suggestion let me know.

kuujo:
Sorry, we need to update this branch now :-(

kuujo:
What I'm guessing is that having incorrect connection information in the state machine prevents it from sending events to the client, and because of the sequential consistency checks on event messages that causes the client to simply keep looping waiting for events that will never come because they're being sent by the wrong server. Does that seem right?

I'm wondering if there's a more elegant solution in that this can be fixed by using a logical timestamp in the client's `ConnectRequest`/`AcceptRequest`/`ConnectEntry` to simply prevent newer connection information from being overwritten by older requests.

It may also be a more fundamental issue that can be fixed by not proactively removing connections in the state machine. I have to poke around a bit and figure out what can/can't be safely done.

kuujo:
Hmm... actually, I think `ConnectRequest`/`ConnectEntry` may no longer even be necessary. They were really a product of when events supported linearizable guarantees, but Madan argued against that and it was ultimately removed. I'm not even sure they serve a critical purpose any more.

kuujo:
As I mentioned above, `ConnectRequest`, `AcceptRequest`, and associated logic appears to no longer be necessary. #275 removes them and significantly simplifies connection management inside the servers. I'm going to close this issue and we should work from that one.

PR #266, opened 2017-01-04
Ensure follower commitIndex is not increased beyond AppendRequest entry index

### LINKED ISSUE #265: Followers may commit inconsistent entries
#261 exposed a case where a leader that becomes a follower can commit an entry from its original term that should have been overridden by entries from the new term.

Here's what happens:
* Server A is elected leader for term 1
* Server A logs commits entries up to index 9, logs an entry at index 10 and then crashes
* Server B is elected leader for term 2
* Server B logs a different entry at index 10 and then commits it via server C
* Server B sends an *empty* `AppendRequest` to server A with `logIndex=9`, `logTerm=1`, and `commitIndex=11` but contains no entries
* The log term/index check passes on server A, but it doesn't truncate its log since no entries were in the request. Server A then commits entry `10` since `commitIndex` is `11`

The follower should have either truncated its log or not increased the commit index beyond the request log index plus new entries so the inconsistent entry was not applied.

kuujo:
This PR fixes #265 by simply following the solution in the Raft paper. The change simply ensures that the follower's `commitIndex` cannot be increased beyond the index of the last entry in an `AppendRequest` to ensure that consistency checks occur prior to committing an entry.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/PassiveState.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/PassiveState.java
@@ -185,8 +185,16 @@ class PassiveState extends ReserveState {
    * Appends entries to the local log.
    */
   protected AppendResponse appendEntries(AppendRequest request) {
+    // Get the last entry index or default to the request log index.
+    long lastEntryIndex = request.logIndex();
+    if (!request.entries().isEmpty()) {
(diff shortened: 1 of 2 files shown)


kuujo:
Had to fix a bug in an unrelated test to get the tests passing.

PR #275, opened 2017-02-01
Remove ConnectRequest/AcceptRequest/ConnectEntry

kuujo:
This is an alternative approach to resolving the issue described in #260

Copycat's `ConnectRequest` and `AcceptRequest` and related protocols were largely left over from when session events had linearizability guarantees. Servers used connection information to associate clients with servers in a distributed fashion so that leaders could proxy events through the appropriate server to the client. That's no longer done. Events are simply sent by the server to which the client is connected when they're applied on that server, so I believe it should be safe to remove `ConnectRequest` and associated codes.

This should solve the issue described in #260 since connections are now based solely on the information provided by Netty. If a client is connected to server x then it's connected to server x. Additionally, connections are now associated with sessions when a `KeepAliveRequest` is sent by the client. Because `KeepAliveRequest`s are sent frequently, that also reduces the chance for future bugs preventing clients from progressing due to stale connection information.


kuujo:
I'm going to close this, replaced by #276

PR #277, opened 2017-02-07
Expire client sessions when command/query fail with unknown session

kuujo:
#276 describes an issue wherein clients don't properly expire their session when they receive an `UnknownSessionException` in response to a command/query. This prevents users of the client API from realistically being able to retry commands/queries across sessions since the client must wait for a `KeepAliveRequest` to fail before recovering their session. This PR adds a check for `UnknownSessionException` in `ClientSubmitter` and proactively expires the session if one is received for any command or query.

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -19,7 +19,7 @@
   <parent>
     <groupId>io.atomix.copycat</groupId>
     <artifactId>copycat-parent</artifactId>
-    <version>1.2.1-rc1</version>
+    <version>1.2.1-SNAPSHOT</version>
   </parent>
 
   <packaging>bundle</packaging>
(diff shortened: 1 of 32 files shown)


jhall11:
Here are the logs from a failed attempt to reconnect an ONOS node to a 3 node cluster. This test would continually restart a node and wait for it to recover, then restart the next, etc. ONOS3 comes back up around 09:58:14,665. It looks like ONOS3's session keeps trying to register. and the leader ONOS2, keeps failing these attempts

 https://drive.google.com/file/d/0BwORWZ1M_qo_d29FaUxBbHg5N2M/view?usp=sharing

kuujo:
Fixed the failed test issue. It was actually a late event in a prior test that was causing later tests to fail, which is why the logs looked so strange.

kuujo:
I think this needs to be merged. The particular issue it solved seems to be resolved, and it's going to start getting mixed with a lot of other bug fixes and complicate my new changelog 😄

PR #279, opened 2017-02-09
Override old sessions when a new session is registered from the same client

kuujo:
This PR improves session management in Copycat servers by proactively expiring past sessions from a client when the client registers a new session. This can occur if the client is partitioned from the cluster and decides to expire its own session. In order to simplify state machines, it's best to ensure that each client can only hold one session at a time.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/ServerSessionManager.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/ServerSessionManager.java
@@ -94,10 +94,14 @@ class ServerSessionManager implements Sessions {
    * Registers a session.
    */
   ServerSessionContext registerSession(ServerSessionContext session) {
+    ServerSessionContext oldSession = clients.remove(session.client());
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/atomix/copycat/server/state/ServerSessionManager.java
```diff
@@ -94,10 +94,14 @@ ServerSessionManager unregisterConnection(Connection connection) {
...
   ServerSessionContext registerSession(ServerSessionContext session) {
+    ServerSessionContext oldSession = clients.remove(session.client());
+    if (oldSession != null) {
+      sessions.remove(oldSession.id());
+    }
     session.setConnection(connections.get(session.client()));
     sessions.put(session.id(), session);
     clients.put(session.client(), session);
-    return session;
+    return oldSession;
```
jhall11: Maybe update the javadoc to include which session is returned and that this can be null?

PR #280, opened 2017-02-11
Travis update to run with both open and oracle JDKs

jhall11:
This PR will change the travis configuration to run 2 parallel jobs, one using oraclejdk8 and one using openjdk8. In order to run using openjdk8, we needed to upgrade to trusty from the default precise.


kuujo:
The build only failed because the output seems to skyrocket on OpenJDK with GC warnings. We might have to silence those warnings when tests are being run.

kuujo:
We should just change those warning to DEBUG or TRACE messages actually. They probably don't make sense as WARNings since Copycat is capable of shutting down a state machine and thus allowing commits to be GCed during normal operation, and that probably shouldn't cause a bunch of warnings. I'll do that for #307 and then we can merge this after that one so this doesn't just break all the builds.

PR #281, opened 2017-02-16
Ensure client/server transports are closed on server leave()

kuujo:
`clientTransport` and `serverTransport` are not properly closed when a server is removed from the cluster. This PR simply adds the same block as is used in `shutdown()` to ensure `Transport`s are properly shut down when the server is.


jhalterman:
👍

PR #286, opened 2017-02-24
Ensure segment lock is flushed to disk prior to deleting compacted segments

kuujo:
This PR fixes a bug that can cause logs to be lost if a node crashes shortly after compaction. Because the segment lock is not flushed to disk prior to deleting compacted segments, it's possible compacted segments could be deleted while an unlocked segment is still on disk, thus causing the server to delete the unlocked segment on restart. Use `getFD().sync()` to ensure the lock is on disk prior to deleting compacted segments.


kuujo:
Actually, this PR seems to be incorrect

PR #309, opened 2017-03-22
Force reset of client connection when connection attempts have been exhausted

kuujo:
This PR is a simple fix to ensure that `ClientConnection` resets connection attempts when clients retry commands. Currently, when a client has exhausted all its connection attempts, commands will repeatedly fail since `connection` is `null` until the next `KeepAliveRequest` is attempted, resetting connection attempts. The way clients retry commands is an inefficiency that will be addressed in a separate PR.


kuujo:
This fix actually messes with some of the failure tests. It doesn't seem like a good fix to be merged. I'm going to close this and implement the fix along with the client protocol changes separately.

PR #310, opened 2017-03-22
Refactor client protocol to more efficiently resolve missing commands

kuujo:
This PR is a refactoring of the protocol with which clients send commands/queries to the cluster. Currently, servers enqueue operations received from clients in memory, and clients use request timeouts to determine that an operation has failed and needs to be resent. This has the effect of causing many operations to time out when one command is lost, and it makes queries — which can't be resent for safety — more likely to fail.

Another interesting issue that was discovered is Java's `CompletableFuture` does not call callbacks in the order in which they're added. For example:

```java
CompletableFuture<Void> future = new CompletableFuture<>();
future.whenComplete((r, e) -> System.out.println("1"));
future.whenComplete((r, e) -> System.out.println("2"));
future.whenComplete((r, e) -> System.out.println("3"));
future.complete(null);
```

prints:

```
3
2
1
```

Copycat clients currently rely on ordering in `CompletableFuture` to submit some operations to the cluster. When the client is connecting to a server, operations that are submitted in parallel with the connection will be sent once the connection is complete. These operations are sent by adding a `whenComplete` callback to the connection future and calling `send` once the connection is complete. However, because `CompletableFuture` calls those callbacks in LIFO order, operations typically arrive at the leader in LIFO order. This means the leader has to resequence those operations, increasing latency and placing undue load on the cluster.

This PR fixes that issue by adding a connect queue to `ClientConnection` to ensure requests are sent in sequential order once the connection is complete.

Finally, `ClientConnection` currently does not reset connection attempts once attempts to all servers have been exhausted. So, if the client attempts to connect to all servers and fails, all future `send` calls on `ClientConnection` will fail until the connection attempts are reset by the next keep-alive. This caused many commands/queries to fail for no reason other than that there wasn't a real attempt to connect and send them. This PR rectifies that issue by ensuring a connection that has exhausted all its connection attempts will reset the attempts when the next operation is sent by the client.


kuujo:
Actually, there's one issue with this PR still. When a new leader is elected, it takes some time for the new leader to learn the client's last completed sequence number. The client has to submit a keep-alive in order for the leader to learn about the correct sequencing for the client. So, when a new leader is elected, the client will resubmit commands until its next keep-alive. Clients should be checking the `lastSequence` number when the leader fails a command and proactively submitting a keep-alive if the client's sequencing has already surpassed that point.

kuujo:
Tests failed due to too much logging in Travis. Rerunning.

kuujo:
@jhall11 approves :-)

PR #313, opened 2017-03-29
introduce session unstability timeout

himanshug:
allows user to set a "unstableTimeout" (disabled by default)

if set, then once session reaches UNSTABLE state.... it will switch to UNSTABLE_PLUS after unstableTimeout , once UNSTABLE_PLUS is reached then DefaultCopycatClient is closed automatically.

@kuujo it is missing docs/unit-tests which I will add if the approach looks ok to you. and, feel free to suggest alternate names for the things introduced.


kuujo:
I think this is a fine approach. I don't love the name `UNSTABLE_PLUS`, but I haven't thought of anything better after a few minutes of thesaurusing :-P

I actually wonder if we should just reuse `EXPIRED` for this. What we're effectively doing is  behaving as if the client's session expired after some duration. In fact, we could conceivably use the negotiated `sessionTimeout` to determine when to proactively expire the client's session, or at least default to the session timeout.

himanshug:
yeah, `UNSTABLE_PLUS` is not great but I couldn't think of a better name :)

I did not want to use `EXPIRED` because, that semantically means that server side explicitly expired the session not the client side, so for `EXPIRED` state, `active` is set to `false`. Keeping `EXPIRED` separate keeps semantics clear.
Also, if we reuse `EXPIRED` and user specified RecoveryStrategies.RECOVER .. then there is not much point as, after prolonged UNSTABLE, we will close the session and get back to retrying again.... OTOH I have a wrapper, which on CopycatClient reaching CLOSED state, recreates the client with new set of server addresses and uses that.

jhall11:
What about `STALE` as a name for the new state?

kuujo:
Better than our ideas :-)

himanshug:
@jhall11 wonder why we couldn't think of `STALE` :)

cool, will update the pr with docs and tests now.

kuujo:
+1

client/src/main/java/io/atomix/copycat/client/session/ClientSessionState.java
```diff
@@ -101,10 +109,28 @@ public long getSessionId() {
...
+        unstableSince = System.currentTimeMillis();
+        return setStateAndCallListeners(state);
+      }
     }
+
+    return this;
+  }
+
+  private ClientSessionState setStateAndCallListeners(Session.State state)
+  {
```
kuujo: Just need to move this bracket up :-)

kuujo:
I think this is great!

client/src/main/java/io/atomix/copycat/client/session/ClientSessionState.java
himanshug: done,

thatz my current intellij formatter , different projects follow different styles :)
I'll need to setup separate formatter for copycat ..... do you guys use intellij ? if yes, can you guys export the formatting spec and put it somewhere?

client/src/main/java/io/atomix/copycat/client/DefaultCopycatClient.java
```diff
@@ -150,6 +154,10 @@ private void onStateChange(Session.State state) {
       case UNSTABLE:
         setState(State.SUSPENDED);
         break;
+      case STALE:
+        setState(State.SUSPENDED);
+        this.close();
```
jhall11: Should we try to recover the session here?

kuujo: Yeah, it seems to me like this state should be invoking the recovery strategy

himanshug: not really, if user wanted recovery to continue happening why would they want to configure the unstabilityTimeout .
If they don't configure unstabilityTimeout then session would never reach STALE state and continue retrying to recover from UNSTABLE state.

kuujo:
Running tests one more time

Issue #307, opened 2017-03-18
Use trace log level for requests/responses


kuujo:
Currently, the `DEBUG` log level is overused. The `TRACE` level should be used for requests and responses. That will greatly cut down on the number of `DEBUG` messages and make some issues easier to debug via the logs.

PR #314, opened 2017-03-30
Decrease log level to TRACE for the majority of log messages

kuujo:
#307


kuujo:
Logs look much nicer :-)

kuujo:
We can also merge #280

PR #321, opened 2017-04-17
always persist <term,vote> metadata on disk

himanshug:
this patch is to enable option (1) described in https://groups.google.com/forum/#!topic/atomixio/v29s36PvhMs .

basically, I would like to be able to store raft log and cluster configuration in memory while <term,vote> metadata should still be persisted on disk for raft safety.

~~there are some TODOs in the PR , I will finish those once we agree on the change.~~

there is intentional duplication in backward compatibility code so that later it could be removed easily.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/storage/Storage.java
+++ b/server/src/main/java/io/atomix/copycat/server/storage/Storage.java
@@ -15,8 +15,8 @@
  */
 package io.atomix.copycat.server.storage;
 
-import io.atomix.catalyst.util.Assert;
 import io.atomix.catalyst.concurrent.ThreadContext;
+import io.atomix.catalyst.util.Assert;
 import io.atomix.copycat.server.storage.snapshot.SnapshotFile;
 import io.atomix.copycat.server.storage.snapshot.SnapshotStore;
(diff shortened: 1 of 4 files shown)


kuujo:
I like it!

himanshug:
@kuujo ok, i updated this PR .. should now be mergeable.

BTW, there is intentional duplication in backward compatibility code so that later it could be removed easily.

himanshug:
@kuujo can this be merge?

kuujo:
Sorry, this will be merged after the next release, which I'm doing right now. Had to cut down on what was merged while we did some testing to find a major performance regression in the 1.2.5 release. That's fixed now, but it's being released on its own so we can get an ONOS release out, and then I'll merge this and it will be in the next release.

Coincidentally, I need this patch! I'm planning to add some in-memory only partitions to ONOS for faster ephemeral data structures, so thanks for the work!

himanshug:
@kuujo thanks, yeah not super urgent but just making sure its not orphaned, glad to see that its useful in other contexts too ... I'm currently setting up the stage in druid side to make discovery pluggable so that we can configure it to use either zookeeper based impl or raft based impl or whatever else. once done we'll deploy a copycat based discovery impl in few sandbox environments.

PR #323, opened 2017-04-21
Reset events on duplicate event indexes

kuujo:


kuujo:
This actually should not be necessary since the server will stream later events anyways.

PR #329, opened 2017-05-04
Ensure only a single connection is created between each server

kuujo:
This PR fixes a bug where multiple connections could be in progress between servers. It simply tracks connection futures in addition to the connections themselves.

### DIFF
--- a/server/src/main/java/io/atomix/copycat/server/state/ConnectionManager.java
+++ b/server/src/main/java/io/atomix/copycat/server/state/ConnectionManager.java
@@ -31,6 +31,7 @@ import java.util.concurrent.CompletableFuture;
 public final class ConnectionManager {
   private final Client client;
   private final Map<Address, Connection> connections = new HashMap<>();
+  private final Map<Address, CompletableFuture<Connection>> connectionFutures = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/atomix/copycat/server/state/ConnectionManager.java
```diff
@@ -31,6 +31,7 @@
 public final class ConnectionManager {
   private final Client client;
   private final Map<Address, Connection> connections = new HashMap<>();
+  private final Map<Address, CompletableFuture<Connection>> futures = new HashMap<>();
```
jhall11: Just my personal preference, so take it or leave it:
Maybe change futures to something more descriptive like "connectionFutures"?

kuujo:
Restarting tests

server/src/main/java/io/atomix/copycat/server/state/ConnectionManager.java
kuujo: I concur. Fixed.

kuujo:
This seems to be the culprit causing Copycat tests to fail so frequently on forks:
https://github.com/travis-ci/travis-ci/issues/3775

PR #331, opened 2017-05-12
Add simple performance test

kuujo:


test/src/main/java/io/atomix/copycat/test/PerformanceTest.java
```diff
@@ -0,0 +1,527 @@
+/*
+ * Copyright 2017-present Open Networking Laboratory
```
jhall11: Not sure if this is the license you want here or not

kuujo: 🤔 I wasn't paying attention to my copyright notice, but I think work I do for ON.Lab is under their copyright? I hate this legal crap...

jhall11: I'm not really sure tbh.

Issue #335, opened 2017-05-22
Under certain disconnection circumstances, the copycat-client-io thread takes 100% CPU


JPWatson:
jstack stack traces alternates between these frames:
```
"copycat-client-io-1" #17 prio=5 os_prio=0 tid=0x00007f5e60755800 nid=0x7412 runnable [0x00007f5e4cd57000]
   java.lang.Thread.State: RUNNABLE
        at java.lang.Throwable.fillInStackTrace(Native Method)
        at java.lang.Throwable.fillInStackTrace(Throwable.java:783)
        - locked <0x00000000f10b3008> (a io.atomix.copycat.session.ClosedSessionException)
        at java.lang.Throwable.<init>(Throwable.java:265)
        at java.lang.Exception.<init>(Exception.java:66)
        at java.lang.RuntimeException.<init>(RuntimeException.java:62)
        at java.lang.IllegalStateException.<init>(IllegalStateException.java:55)
        at io.atomix.copycat.session.ClosedSessionException.<init>(ClosedSessionException.java:29)
        at io.atomix.copycat.client.session.ClientSessionSubmitter.submit(ClientSessionSubmitter.java:144)
        at io.atomix.copycat.client.session.ClientSessionSubmitter.access$300(ClientSessionSubmitter.java:51)
        at io.atomix.copycat.client.session.ClientSessionSubmitter$CommandAttempt.lambda$fail$0(ClientSessionSubmitter.java:370)
        at io.atomix.copycat.client.session.ClientSessionSubmitter$CommandAttempt$$Lambda$109/2008009084.run(Unknown Source)
        at io.atomix.catalyst.concurrent.Runnables.lambda$logFailure$0(Runnables.java:20)
        at io.atomix.catalyst.concurrent.Runnables$$Lambda$31/1944702768.run(Unknown Source)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #336, opened 2017-05-25
Prevent recursion of commands during closed session state

kuujo:
This PR fixes a bug that appears to lead to infinite (asynchronous) recursion in the client. When a session is closed and its state is set to `EXPIRED` or `CLOSED`, additional operations submitted to the session will enter an infinite loop because `ClosedSessionException` is not properly handled in the `CommandAttempt.fail` method. This PR simply checks to ensure the exception is not a `ClosedSessionException` and effectively breaks out of the loop if it is.


JPWatson:
do you think this might fix https://github.com/atomix/copycat/issues/335?

PR #339, opened 2017-07-12
Make number of KeepAlives Configurable

JPWatson:
We're seeing sessions regularly becoming expired. 
Would like to be able to be able to increase the number of keep alives to see whether this helps.


jhall11:
Did increasing the number of timeouts have help session stability in your environment?

